PR #56, opened 2017-05-25
Test for nonmanifold surfaces and small bug fix

raovgarimella:
This adds a test to ensure that we can read and process a mesh/model of non-manifold surfaces (in this case two planar surfaces meeting at a right angle in 3-space).


lipnikov:
No conflicts on serrano1 with TPLS 0.94.1. Recommend merging.

PR #63, opened 2017-06-01
Changes to allow faces to be connected to more than 2 cells

raovgarimella:
The primary changes here are to allow for non-manifold surfaces where multiple surfaces may intersect along an edge in 3-space. This requires that "faces" be allowed to connect to more than 2 "cells".

Bundled in (should have been separate) is the fix to avoid inheriting labeled sets from which surface meshes are extracted

### DIFF
--- a/src/mesh/Mesh.cc
+++ b/src/mesh/Mesh.cc
@@ -419,24 +419,22 @@ Mesh::compute_face_geometric_quantities_() const
 
   face_areas_.resize(nfaces);
   face_centroids_.resize(nfaces);
-  face_normal0_.resize(nfaces);
-  face_normal1_.resize(nfaces);
+  face_normals_.resize(nfaces);
 
   for (int i = 0; i < nfaces; i++) {
     double area;
-    AmanziGeometry::Point centroid(space_dim_), normal0(space_dim_),
-        normal1(space_dim_);
+    AmanziGeometry::Point centroid(space_dim_);
+    std::vector<AmanziGeometry::Point> normals;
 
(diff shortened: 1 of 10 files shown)


ecoon:
The only potential issue I see here is that sometimes it DOES do what we expect, and we rely on it to do so.

In the case that a set of faces intersects with the manifold/lifted mesh exactly as a set of edges, then the face set was lifted, and was properly referred to as a face set (because the face set became edges, and those "edges" are "faces" in a 2D submanifold.

This is important because we used it for boundary conditions on the surface mesh.  For instance, say you have a really irregular mesh (a watershed) in which you have tagged all "side boundary faces" (not in the bottom or top boundary).  Then when you lift the "surface mesh", the "side boundary faces" set becomes a new "side boundary faces" face set on the surface mesh consisting of the intersection, which is exactly the 1D boundary of the new surface mesh.  This can then be used to specify BCs for surface flow problems.

This use case still needs to be allowed.  Not sure if it is or not.

jd-moulton:
It seems we need these two issues to be separated so we can move forward on them at different paces. 

Rao please comment further on Ethan's question.   I think the primary changes that are here could be accepted (faces connected to more than two cells), but we seem to need requirements/understanding to be reached on facesets from surfaces.

raovgarimella:
Ethan and I talked about it. What is here works - it does not break any thing and actually avoids doing the wrong thing. Once this PR is in, I will put in the fix to do the right thing (where the face set from which we are creating a surface mesh gets inherited as a cell set).

PR #69, opened 2017-06-15
Fixes #67

### LINKED ISSUE #67: Test to collapse degeneracies fails MeshAudit checks in parallel
A test to collapse out degeneracies fails a MeshAudit check in parallel. The specific failure is that the owned face global IDs are not contiguous. This can obviously happen due to the mesh modification and I don't see that the mesh is being renumbered.

It was not detected earlier because @ekikinzon and @ecoon were using a standalone degeneracy eliminator app and then pulling into Amanzi

ekikinzon:


src/mesh/mesh_mstk/test/test_pinchouts.cc
```diff
@@ -0,0 +1,196 @@
...
+
+#include "../../mesh_factory/MeshFactory.hh"
+#include "../Mesh_MSTK.hh"
+
+
+int main(int argc, char *argv[])
+{
+  MPI_Init(&argc, &argv);
+  
+  std::string xml_filename = "test/po_test_pri.xml";
```
raovgarimella: Unfortunately this filename is still hardcoded which means only the prism mesh or the hex mesh will get tested. You can do one of two things - make this a "test fixture" and call it for both cases (see https://github.com/unittest-cpp/unittest-cpp/wiki/Writing-More-Tests-With-the-Bowling-Game-Kata-Test-Two) or make it take an argument and somehow run the test with different arguments

raovgarimella:
Also, think of a way we can make this a standalone app that can be run on any mesh so that we can test it before setting up an actual simulation

src/mesh/mesh_mstk/Mesh_MSTK.cc
```diff
@@ -5536,6 +5604,26 @@ void Mesh_MSTK::inherit_labeled_sets(MAttrib_ptr copyatt)
       if (!mset_parent)
         continue;
 
+      // Also, if this is a lower dimensional mesh (like a surface
```
raovgarimella: Lets take this out of this PR

src/mesh/mesh_mstk/Mesh_MSTK.cc
```diff
@@ -4687,15 +4692,6 @@ void Mesh_MSTK::collapse_degen_edges()
 
       /* Degenerate edge  - must collapse */
 
-      /* If its the first time, we have to allocate
-         these lists */
-      if (!entities_deleted) {
-        deleted_vertices = List_New(0);
```
raovgarimella: @jd-moulton wants to keep the old and new versions of the code. Lets wrap it in a #defined statement like 

#ifdef (MSTK_3_0_0_or_NEWER)
.....new code....
#else 
.....old code....
#endif

src/mesh/mesh_mstk/test/test_pinchouts.cc
lipnikov: Creating tests with input arguments is easy, see  e.g. operator_diffusion_nlfv.cc

PR #96, opened 2017-08-23
Fix for Boost TPL picking a wrong compiler on OS X if Homebrew gcc is…

ekikinzon:
… used.

Also makes sure that the correct toolset is used.


jd-moulton:
After discussing the update with Eugene, it appears that there is an issue with previous way we were setting the toolset (i.e., toolset=gcc-${CMAKE_CXX_COMPILER_VERSION}), and Eugene's addition provides a fallback that works.  It's likely what was happening on most systems anyway.  This needs a little more investigation to resolve the ambiguity.

PR #100, opened 2017-08-30
nem_slice patch to make columns reside on single partition

raovgarimella:
I talked to Greg Sjaardema from Sandia and he is willing to consider putting this in the source directly


lipnikov:
Failed 

in applications/nem_slice/fix_column_partitions.C:191:25: error: there are no arguments to fabs that depend on a template parameter, so a declaration of fabs must be available.

Use include <cmath>

lipnikov:
Build passed on Mac.

PR #122, opened 2017-09-25
Add partitioner specification to mesh factory

raovgarimella:
Added a method to mesh factory that allows callers to specify the partitioner to be used. Default partitioner is METIS. Changes allow us to reinstate test_deform_vols and test_columns unit tests as we can now specify that these must use ZOLTAN_RCB as the partitioner.


raovgarimella:
@lipnikov suggested that the addition of partitioner to mesh_factory be propagated up to the input spec level. The latest set of commits does that.

raovgarimella:
Updated v2.3.1-draft documentation

raovgarimella:
BTW, @jd-moulton, the nemslice patch I sent you is supposed to be a replacement for the original patch, not an add-on

ekikinzon:
Konstantin has already replaced the old patch by the new one (and also ran tests with it included).

lipnikov:
Works for me.  Need on a feature branch.

PR #126, opened 2017-10-05
Added some unit tests for degeneracy elimination

raovgarimella:


raovgarimella:
I am going to merge this because the next set of commits will be a big shift (internally) in where we do the degeneracy elimination. We were doing it as a post-processing step after partitioning or reading it as a pre-partitioned mesh. However, it was too hard to keep the changes consistent across processor boundaries.

In the next version we will eliminate degeneracies in a "serial" manner on meshes with no ghosts and then establish the parallel mesh. More on that in the next PR.

PR #134, opened 2017-10-26
Moved collapse degenerate edges routine to earlier in call sequence

raovgarimella:
Collapsing degenerate edges and eliminating degenerate cells while keeping everything consistent is messy for distributed meshes. Therefore, we moved the call to the routine earlier in the call sequence so that the process is more serial in nature. This should make degeneracy elimination in parallel runs more robust.

In the case where we read a mesh on one processor and partition/distribute to all processors, degeneracy elimination is now done before partitioning and distribution rather than after. This does add extra load on P0.

In the case where we read multiple pre-partitioned Nemesis files (with global IDs indicating the relationship between entities of different partitions), we do the degeneracy elimination using a consistent convention on each processor before we aim to build up ghost layers and establish interprocessor connectivity.


raovgarimella:
So, I have verified that this version of the code (along with parallel changes in MSTK to be tagged soon as 3.0.2)  is able to read a coarse san juan river prepartitioned mesh for 16 and 64 processor cases. I also tried it for a finer San Juan mesh for 16 processors.

PR #137, opened 2017-11-17
adds support to Mesh_MSTK for RegionType::ALL, and tests in serial and parallel

ecoon:
I spent some time trying to find how this was implemented, as I'm fairly sure it was at some point.  Maybe in MSTK proper?  Even if it is implemented there, I think this is the better strategy.

The RegionType::ALL was introduced a while ago to allow MeshLogical to use regions effectively.  This adds support for this region type to Mesh_MSTK in addition to MeshLogical.

I like the semantics of this because it requires the user to still specify the region, with a "region: all" type.  This means there is no "magic name" to remember (or worse, mistakenly use that name to refer to a different region type).

### DIFF
--- a/doc/input_spec/AmanziNativeSpecV7.rst
+++ b/doc/input_spec/AmanziNativeSpecV7.rst
@@ -4829,7 +4829,7 @@ User-defined regions are constructed using the following syntax
    * Shape [list] Geometric model primitive, choose exactly one of the following: 
      `"region: point`", `"region: box`", `"region: plane`", `"region: labeled set`", 
      `"region: layer`", `"region: surface`", `"region: boundary`",
-     `"region: box volume fractions`", or `"region: line segment"`.
+     `"region: box volume fractions`", `"region: line segment"`, or `"region: all"`.
 
 Amanzi supports parameterized forms for a number of analytic shapes, as well as more complex 
 definitions based on triangulated surface files.  
@@ -5123,6 +5123,20 @@ points.
     </ParameterList>     
 
 
+All
+....................
+
+List *region: all* desribes a region which matches all entities in the
+mesh.  No parameters are required.
+
+.. code-block:: xml
+
+   <ParameterList name="ENTIRE MESH"> <!-- parent list -->
+      <ParameterList name="region: all">
+      </ParameterList>
+    </ParameterList>     
+    
+
 Notes and example
 -----------------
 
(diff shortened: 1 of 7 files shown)


ecoon:
Added Konstantin as a reviewer simply so that he knows that this new capability is then available.

lipnikov:
1. Your tests do not cover the case of nodes or faces of region ALL.

2. We already support
<ParameterList name="region: box">
    <Parameter name="low coordinate" type="Array(double)" value="{-9.99e+98, -9.99e+98}"/>
    <Parameter name="high coordinate" type="Array(double)" value="{9.99e+98, 9.99e+98}"/>
</ParameterList>

This region can extract cells, faces or nodes. Your solution is longer if both cells and faces are needed by an algorithm.  Although I do not see a problem with having an additional way to set up a region, I'd like to understand more clearly if there exists a use case when region:all would be better than region::box.

raovgarimella:
So, I think it was implemented in the input translator because we did not want to have the burden of an entity set that contained all the mesh entities. This was discussed and resolved in 2014 in ticket number 1293 on the old software.lanl.gov Trac system. I wonder how we can retrieve it.

ecoon:
The example use not covered by "region: box" is not in an MSTK mesh, but in a MeshLogical, which need not have centroids geometrically defined.  For instance, cell centroids may have the same (x,y,z), but differ in some other, parametric value, e.g. a multiple interacting continua porosity model, where every geometric cell has a 1D "mesh" hanging off of them.  The result of this type of "mesh" is that centroids may not need to exist -- cell volumes, face areas, and perpendicular bisectors are all relative quantities and are sufficient to define a PDE.  Often in this case we define a "convenient" (x,y,z) centroid that helps make plotting easier, but is not tied to a geometric model.

This is what motivated "region: all" in the first place.  Extending it to be supported in MSTK is only to allow uniformity across input files.

I thought the tests did cover face and node checks -- let me look again, maybe I missed something.

What do you mean by "longer"?  Less performant?  In memory use or computation time?  That somewhat surprises me since "all" forgoes the "inside" check needed for "box" which should be relatively expensive.

ecoon:
I'm confused -- which part of the tests don't check FACE and NODE?  See, for instance, starting line 87 in src/mesh/mesh_mstk/test/test_hex_3x3x3_sets.cc

raovgarimella:
If we are going to use a large box to scoop up all mesh entities, I would rather have Region:All. However, before I chime in for or against, I want to see the old ticket on the Trac site to remind myself of what we discussed. Dave said he will try to resurrect it.

lipnikov:
Re:  line 87 in src/mesh/mesh_mstk/test/test_hex_3x3x3_sets.cc
Okay I missed this. I've noticed that .xml files have only cells.

lipnikov:
Re: What do you mean by "longer"? 
I meant to say that a user must to create more regions (longer input spec) when both faces and cells are required by an algorithm.

lipnikov:
No more comments. I think this PR is ready to merging.

jd-moulton:
I setup the old Trac site locally, and printed Ticket 1293 out as a PDF.

[Trac - Ticket #1293.pdf](https://github.com/amanzi/amanzi/files/1497836/Trac.-.Ticket.1293.pdf)

Conversation is similar, good to have to review.  Seems like we can move ahead with this proposal, so I'll merge it.

raovgarimella:
Looks good.

PR #140, opened 2017-11-25
Konstantin/tpls

lipnikov:
Moved from configure to cmake for NetCDF.


jd-moulton:
I  had a problem with the project() command overwriting the CMAKE compiler flags we had defined, so I moved it earlier. Also fixed an issue with SILO, which wasn't building shared libraries for me.  Please test and let me know if this works for you.

AlexisPerry:
Two things...  First, boostrap.sh triggers an openmpi build, even though I have an openmpi module loaded.  If I specify --with-mpi=<dir>, then it uses the loaded module.  Second, the TPLs build fine, but Amanzi itself fails with the following:
...
Scanning dependencies of target tPPH
[ 51%] Building CXX object src/structured_grid/utils/CMakeFiles/tPPH.dir/tParmParseHelpers.cpp.o
[ 51%] Building CXX object src/state/CMakeFiles/state.dir/Field_Scalar.cc.o
[ 52%] Building Fortran object src/structured_grid/utils/CMakeFiles/tPPH.dir/__/shared_dso.f90.o
[ 52%] Linking CXX executable tPPH
/usr/bin/ld: cannot find -lgeochemistry
collect2: error: ld returned 1 exit status
make[2]: *** [src/structured_grid/utils/tPPH] Error 1
...

lipnikov:
Thanks for testing. The current design is to rely solely on --with-mpi and not on the loaded modules.
If --with-mpi is not specified, openmpi will be build. This gives more control to a developer. Two examples are provided with ./bootshrap.sh --help

lipnikov:
Multiple tests passed on OSX and Ubuntu. The build looks stable to me. Default behaviour can be easily changed in the future with minor tweaking of bootsrap.sh.

lipnikov:
MOAB issues:
1. misses some CMakeList.txt in recent release 5.0.0 
2. search for Zoltan is not consistent with Trillions installation name
Should wait for new release.

lipnikov:
Merged approved by PI.

PR #146, opened 2017-12-08
Ats coupled chem

dasvyat:
Main changes:

multidomain names of chemistry fields.
Ability to consider density as a MultiVector


lipnikov:
Alquimia_PK.cc:
We typically use 4 spaces after ":". 
I think that default constructor of the base class is called automatically, no need to repeat it.
Remove comments (around lines 326 and 850)
Is convert2mole_fraction_  used? 

Amanzi_PK.cc:
double water_density = 998.2. water_density=S->GetScalarData(fluid_den_key_) is more reliable, since a user may initialize it at a different temperature. 

Amanzi_PK.hh:
CopyFieldstoNewState could be inherited from the base class. 

Field.hh:
Remove comments form destructor.

State.cc:
No need to change this file. Remove empty lines.

PR #165, opened 2018-03-02
Renamed USED to ALL and added namespace protecion of names. closes #164

### LINKED ISSUE #164: Replaced USED with a more clear name
Suggestion was to replace USED with ALL and protect names using namespace (enum class).

lipnikov:
Renamed USED to ALL.
Replaced enum Parallel_type with enum class Parallel_type.

### DIFF
--- a/src/mesh/MeshDefs.hh
+++ b/src/mesh/MeshDefs.hh
@@ -102,20 +102,18 @@ std::string entity_kind_string(Entity_kind kind)
 }
 
 // Parallel status of entity 
-    
-enum Parallel_type 
-{
-  PTYPE_UNKNOWN = 0, // Initializer
-  OWNED = 1,         // Owned by this processor
-  GHOST = 2,         // Owned by another processor
-  USED  = 3          // OWNED + GHOST
+enum class Parallel_type {
+  PTYPE_UNKNOWN = 0,
+  OWNED = 1,  // Owned by this processor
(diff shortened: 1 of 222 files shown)


src/mesh/MeshDefs.hh
```diff
@@ -102,20 +102,18 @@ std::string entity_kind_string(Entity_kind kind)
...
 
 // Parallel status of entity 
-    
-enum Parallel_type 
-{
-  PTYPE_UNKNOWN = 0, // Initializer
-  OWNED = 1,         // Owned by this processor
-  GHOST = 2,         // Owned by another processor
-  USED  = 3          // OWNED + GHOST
+enum class Parallel_type {
```
raovgarimella: If we ever change to Jali, we will have to rename this enum class to ```Entity_type```. Jali has ```PARALLEL_OWNED, PARALLEL_GHOST, BOUNDARY_GHOST, ALL``` and so I decided that the name ```Parallel_type``` wouldn't fit.

lipnikov: type and kind are often interchangeable. Hence Entity_kind and Entity_type lead to confusion.
Parallel_type is the  more informative name.

lipnikov:
BTW, "enum class" revealed a few oops in CompositeVectorFunction using a silent conversion of enum to bool.

PR #186, opened 2018-05-07
Konstantin/dg

lipnikov:
This PR does not change interfaces to high-level structures such as  PKs and Operators. It is for informational purposes only. Comments and suggestions from team members are welcome and will be greatly appreciated. This PR will be merged to master in 1 week from its creation.

1. Restructured DG_Modal class by extracting three DG bases in separate classes. Available bases are "natural", "normalized" and "orthonormalized". Most calculations are performed in the natural basis which is stable to round-off errors. Then, matrices and vectors are transformed to the selected basis. 

2. Enriched content of analytic solutions for testing DG schemes. Now, an analytic solution satisfies this equation:  a du/dt + v . grad(u) - div(K grad(u)) + r u = f

3. Added boundary conditions for high-order transport. Added unit tests for them with cubic polynomial approximations and RK3 time integration schemes.

4. Code cleaning. Default implementation of a few virtual functions improved readability of derived classes.  Removed excessive use of keyword "virtual" inside derived classes in WhetStone. Made projectors the required members of MFD classes (work in progress). 

5. Added new mesh map to WhetStone and created a factory of mesh maps. This capability will be used for data transfer between meshes.

6. Added new tools to the numerical integration capability of WhetStone. At the moment, it has various algorithms for integrating polynomials over polytopal cells.

7. Updated operator's section of the native documentation.

8. Added support of direct solvers. Fixed bugs in matrix assembly do dG schemes revealed by these solvers.


AlexisPerry:
Once Dave has the new container built, go to travis-ci.org and restart the build for the PR to make sure it passes.  If it succeeds, then in my opinion that's all you need.

PR #195, opened 2018-05-24
better enables "systems" in Hypre BoomerAMG

ecoon:
Hypre has a "systems" of equations approach in which, through a variety of flags, one can specify to coarsen the operator not as a single set of unknowns, but independently or collectively as block systems of equations.  This is critical in cases where scaling between the systems is not perfect or nondimensional.  Previously we allowed this through the "number of functions" option in PreconditionerBoomerAMG -- this specified a fixed, strided block operator structure where, for instance, if there were 2 "functions", then structure was assumed to be "1,2,1,2,1,2..." for each unknown.  

Unfortunately, while straightforward, this approach did not allow splitting face and cell unknowns for independent coarsening.  This pull request automates the generation of a "block index" array, which can be passed to Hypre to specify which block each unknown belongs to for coarsening.  

This array of ints can be provided by SuperMap, and is currently set up to provide a unique index for each component and each degree of freedom in the SuperMap.  For hybrid form discretizations, this means faces and cells can be coarsened independently.  For, e.g., coupled flow + energy problems, this means 4 unique block ids (face + cell x flow + energy).  

Additionally, to avoid repeating some work, the Operator InitPreconditioner() was extended to include an InitializePreconditioner() and an UpdatePreconditioner().  The InitializePreconditioner() would then typically be called once (for each SymbolicAssembleMatrix() call) while the UpdatePreconditioner() should be called each time the Matrix entries are changed (after each AssembleMatrix() call).  This is different from the previous interface where InitPreconditioner() needed to be called every time the matrix entries changed.

Note the InitPreconditioner() still exists, and does not generate block IDs.  So this pull request does not change answers.  However, this should be deprecated, and PKs can swap to the Initialize() / Update() model instead, as this is always not less efficient.  When they do so, they will also get the updates to how BoomerAMG is used.  Additionally, other Preconditioner* implementations could take better advantage of this splitting to be more efficient.

Note that I have tried this face + cell split coarsening on 4 ATS meshes (1 quasi-2D hillslope + 3x 3D triangular prisms at a variety of scales and DEM nastiness).  In these experiments, I solved problems of coupled surface/subsurface water, and saw improved convergence rates on all DEM-based meshes, and not reduced convergence rates on simpler meshes.  I believe it is justified making this the default for all problems using BoomerAMG.  If people want to do some testing, I could amend this pull request to add an option enabling this method.

### DIFF
--- a/src/data_structures/SuperMap.cc
+++ b/src/data_structures/SuperMap.cc
@@ -129,13 +129,13 @@ SuperMap::GhostIndices(const std::string& compname, int dofnum) const {
 }
 
 
-std::pair<int, Teuchos::RCP<std::vector<int>>>
+std::pair<int, Teuchos::RCP<std::vector<int> > >
 SuperMap::BlockIndices() const {
   auto block_indices = Teuchos::rcp(new std::vector<int>(Map()->NumMyElements()));
   int block_id = 0;
   for (const auto& comp : *this) {
     int ndofs = NumDofs(comp);
-    for (int d = 0; d!=ndofs; ++d) {
+    for (int d=0; d!=ndofs; ++d) {
       const auto& inds = Indices(comp, d);
       for (int i=0; i!=inds.size(); ++i) (*block_indices)[inds[i]] = block_id;
       block_id++;
--- a/src/data_structures/SuperMap.cc
+++ b/src/data_structures/SuperMap.cc
@@ -129,6 +129,22 @@ SuperMap::GhostIndices(const std::string& compname, int dofnum) const {
 }
 
 
+std::pair<int, Teuchos::RCP<std::vector<int>>>
+SuperMap::BlockIndices() const {
+  auto block_indices = Teuchos::rcp(new std::vector<int>(Map()->NumMyElements()));
+  int block_id = 0;
+  for (const auto& comp : *this) {
+    int ndofs = NumDofs(comp);
+    for (int d = 0; d!=ndofs; ++d) {
+      const auto& inds = Indices(comp, d);
+      for (int i=0; i!=inds.size(); ++i) (*block_indices)[inds[i]] = block_id;
+      block_id++;
+    }
+  }
+  return std::make_pair(block_id, block_indices);
+}
+
+
 const std::vector<int>&
 SuperMap::CreateIndices_(const std::string& compname, int dofnum, bool ghosted) const {
   if (ghosted) {
--- a/src/data_structures/SuperMap.hh
+++ b/src/data_structures/SuperMap.hh
@@ -67,7 +67,7 @@ class SuperMap {
   // where each dof and component have a unique integer value.  The returned
   // int is the number of unique values, equal to
   // sum(NumDofs(comp) for comp in components), in this array.
-  std::pair<int, Teuchos::RCP<std::vector<int>>> BlockIndices() const;
+  std::pair<int, Teuchos::RCP<std::vector<int> > > BlockIndices() const;
   
   // iterate over compnames
   typedef std::vector<std::string>::const_iterator name_iterator;
--- a/src/data_structures/SuperMap.hh
+++ b/src/data_structures/SuperMap.hh
@@ -63,6 +63,19 @@ class SuperMap {
   const std::vector<int>& Indices(const std::string& compname, int dofnum) const;
   const std::vector<int>& GhostIndices(const std::string& compname, int dofnum) const;
 
+  // block indices.  This is an array of integers, length Map().MyLength(),
+  // where each dof and component have a unique integer value.  The returned
+  // int is the number of unique values, equal to
+  // sum(NumDofs(comp) for comp in components), in this array.
+  std::pair<int, Teuchos::RCP<std::vector<int>>> BlockIndices() const;
+  
+  // iterate over compnames
+  typedef std::vector<std::string>::const_iterator name_iterator;
+  name_iterator begin() const { return compnames_.begin(); }
+  name_iterator end() const { return compnames_.end(); }
+  unsigned int size() const { return compnames_.size(); }
+
+  
   // // block accessors
   // void BlockIndices(const std::string& compname, int element_lid, std::vector<int>* indices) const {
   //   int ndofs = NumDofs(compname);
--- a/src/operators/Operator.cc
+++ b/src/operators/Operator.cc
@@ -21,6 +21,7 @@
 #include "Epetra_Vector.h"
 
 // Amanzi
+#include "dbc.hh"
 #include "DenseVector.hh"
 #include "MatrixFE.hh"
 #include "PreconditionerFactory.hh"
@@ -380,9 +381,8 @@ void Operator::InitPreconditioner(const std::string& prec_name, const Teuchos::P
 {
   AmanziPreconditioners::PreconditionerFactory factory;
   preconditioner_ = factory.Create(prec_name, plist);
-  preconditioner_->Update(A_);
+  UpdatePreconditioner();
 }
-
 
 /* ******************************************************************
 * Initialization of the preconditioner. Note that boundary conditions
@@ -394,6 +394,47 @@ void Operator::InitPreconditioner(Teuchos::ParameterList& plist)
   preconditioner_ = factory.Create(plist);
   preconditioner_->Update(A_);
 }
+
+
+/* ******************************************************************
+* Two-stage initialization of preconditioner, part 1.
+* Create the PC and set options.  SymbolicAssemble() must have been called.
+****************************************************************** */
+void Operator::InitializePreconditioner(Teuchos::ParameterList& plist)
+{
+  AMANZI_ASSERT(A_.get());
+  AMANZI_ASSERT(smap_.get());
+
+  // provide block ids for block strategies.
+  if (plist.isParameter("preconditioner type") &&
+      plist.get<std::string>("preconditioner type") == "boomer amg" &&
+      plist.isSublist("boomer amg parameters")) {
+
+    // NOTE: Hypre frees this
+    auto block_ids = smap_->BlockIndices();
+
+    plist.sublist("boomer amg parameters").set("number of functions", block_ids.first);
+
+    // Note, this passes a raw pointer through a ParameterList.  I was surprised
+    // this worked too, but ParameterList is a boost::any at heart... --etc
+    plist.sublist("boomer amg parameters").set("block indices", block_ids.second);
+  }
+
+  AmanziPreconditioners::PreconditionerFactory factory;
+  preconditioner_ = factory.Create(plist);
+}
+
+/* ******************************************************************
+* Two-stage initialization of preconditioner, part 2.
+* Set the matrix in the preconditioner.  Assemble() must have been called.
+****************************************************************** */
+void Operator::UpdatePreconditioner()
+{
+  AMANZI_ASSERT(preconditioner_.get());
+  AMANZI_ASSERT(A_.get());
+  preconditioner_->Update(A_);
+}
+
 
 
 /* ******************************************************************
--- a/src/operators/Operator.cc
+++ b/src/operators/Operator.cc
@@ -413,7 +413,7 @@ void Operator::InitializePreconditioner(Teuchos::ParameterList& plist)
     // NOTE: Hypre frees this
     auto block_ids = smap_->BlockIndices();
 
-    plist.sublist("boomer amg parameters").set("number of functions", block_ids.first);
+    plist.sublist("boomer amg parameters").set("number of unique block indices", block_ids.first);
 
     // Note, this passes a raw pointer through a ParameterList.  I was surprised
     // this worked too, but ParameterList is a boost::any at heart... --etc
--- a/src/operators/Operator.hh
+++ b/src/operators/Operator.hh
@@ -220,6 +220,8 @@ class Operator {
 
   void InitPreconditioner(const std::string& prec_name, const Teuchos::ParameterList& plist);
   void InitPreconditioner(Teuchos::ParameterList& plist);
+  void InitializePreconditioner(Teuchos::ParameterList& plist);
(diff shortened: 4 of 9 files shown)


lipnikov:
1.  Have you tested the new code with Hypre 2.14.0 which is Amanzi's new standard? My approval is contingent to positive answer to this question.

2. It is good to have finer control of preconditioners. But I am curious: How expensive is InitializePreconditioner? For ILU and AMG lots of work go into building decomposition, smoothers, etc, that are parts of UpdatePreconditioner.

3. Non-consistent syntax in if-else statements. "d = 0" in one loop and "i=0" in the next one. We typically use space between brackets ">>" in <std::vector<int>>. Although space is no longer needed, for code consistency we have to pick one style.

jd-moulton:
This sounds great, I didn't realize that Hypre had this flexibility built in.   I like that there is a path to migrate the PKs to the new split InitializePreconditioner() and an UpdatePreconditioner().   This seems like a potentially more powerful model in general.   There aren't a lot of PKs being added right now so I'm curious if people have thoughts no how long we should take to make this fairly simple transition (a week, a month?).   

Please add an option to control this behavior though, as you suggested you could. Hypre updates, and other comparisons or quirks may result in us wanting to compare the performance of old and new approaches.   We will definitely want to do some testing on things like F-Area and DVZ for dry infiltration conditions and probably Daniil has some other Arctic examples. 

What are your thoughts on adding this option to the stable 0.87 branch (as a patch update tagged at say 0.87.1).  Or do you want to push for a more complete update to 0.88 soon?

PS.  Added Daniil as a reviewer as he's active in solver related parts of the code.

ecoon:
1. I'm working on testing this on the Hypre 2.14, but see Issue #199.  I'll also set up to test this on operators_diffusion_curved, which should work now.

2. a) Currently Ifpack_Hypre does a lot of work making Hypre vectors, matrices, etc, to pass in the data to Hypre.  These things are now done in InitializePreconditioner(), and it is premature pessimization to free and re-alloc these things each iteration.  It's probably not a huge fraction of the time (compared to allocing the whole coarse-grid heirarchy) but it certainly isn't free. 
2. b) HYPRE does NOT allow you to re-use the coarse-grid heirachy -- this is always done repeatedly in every call to UpdatePreconditioner().  This is a Hypre limitation, not an Ifpack one.
2. c) To me, the key is that we shouldn't be re-constructing the Preconditioner object each Update call.  This is a design limitation as it limits how individual Preconditioner objects can try to hide work through reuse.  How well each individual Preconditioner implementation takes advantage of it is up to the PC.  We already have this pattern in the Preconditioner objects (they have an Init() and an Update()), but I would argue this is a design limitation in our Operator class (which always re-allocates the Preconditioner object each InitPreconditioner() call).

3. On time to refactor PKs, I did all the ATS ones in about 10 minutes.  Effectively it just means adding InitializePreconditioner() calls everywhere there are SymbolicAssemble() calls, and changing InitPreconditioner() calls to UpdatePreconditioner().  Then a critical eye should look at each Preconditioner* object and see if work done in Update() can be done in Init().  But I bet those are already reasonably selected.

4. Will add the option.

ecoon:
Looks like the Travis CI run needs to get its time limit extended?

ecoon:
I've updated the pull request to:

1. add and document an option "use block indices" to PreconditionerBoomerAMG, defaulting to False, which toggles whether to use this strategy.  Note that setting this option does not mean that existing PKs will use it -- the indices must also be provided by calling the right Initialize/Update paradigm.  Using the option without updating the PK code will currently throw.

2. I've tested it on ATS runs that suggested the change and with Hypre 2.14, and it does not change the answer (that this never hurts and, at least for the handful of meshes I've tried that were all triangular prisms with bad aspect ratios, helps significantly). 

3. On operators_diffusion_curved, it decreases the number of iterations by 3.  Which is nothing against the 346-or-so iterations it currently takes.  But it doesn't hurt either.  Neither of these things should be surprising -- it is easy to see if a run will benefit from this change by looking at a Verbose output of Boomer.  I'll post this info once a few more runs are done, but the short of it is that coarsening resulted in very bad row-sums without it.

4.  Add some spaces, deleted some others.  If you actually want to be this conforming on style, please provided a clang-format style sheet to automate this process and let's be done with it.

ecoon:
So there are some fundamental issues in the Ifpack implementation of Hypre that don't follow their own documentation.  Specifically, Initialize() is supposed to depend upon the matrix sparsity structure but not the values, while Compute() is supposed to depend upon the values.  Their implementation, however, only assembles the Hypre operator in Initialize().  The result is that we've been completely re-generating a lot of Hypre vectors, matrices, etc every nonlinear iteration, when we really shouldn't need to.  Again, note this is the global matrices and vectors, not the coarsened ones, which really do need to be regenerated every time the _values_ change.  But I bet this is a significant cost.

Does anyone have a strong opinion on whether to re-implement Ifpack_Hypre or to bypass Ifpack in PreconditionerBoomerAMG?  Really, maybe the "right" thing to do is have a PreconditionerIfpack() which should generalize this, then re-implement Ifpack_Hypre and contribute it back to Trilinos...

jd-moulton:
What does this look like in Ifpack2.  My impression is that we should be moving there at some point, maybe some of these issues are resolved there.

ecoon:
For what it's worth, here is an example output of a Boomer AMG case where Boomer fails without the block structure.  As you can see, the interpolation rowsums are very bad, and the V-cycle actually increases the residual:

```
BoomerAMG SETUP PARAMETERS:

 Max levels = 25
 Num levels = 11

 Strength Threshold = 0.850000
 Interpolation Truncation Factor = 0.000000
 Maximum Row Sum Threshold for Dependency Weakening = 0.500000

 Coarsening Type = Falgout-CLJP
 measures are determined locally
 No global partition option chosen.
 Interpolation = extended+i interpolation
Operator Matrix Information:

            nonzero         entries per row        row sums
lev   rows  entries  sparse  min  max   avg       min         max
===================================================================
 0  303447  3276583  0.000     6   11  10.8  -2.127e+01   1.948e+01
 1  102984  1649528  0.000     5   26  16.0  -5.006e+12   1.813e+30
 2   50709   958381  0.000     5   35  18.9  -5.091e+00   9.347e+01
 3   22446   393988  0.001     3   37  17.6  -5.108e+00   1.151e+02
 4    8132   131950  0.002     1   37  16.2  -4.560e+00   1.615e+01
 5    3055    48187  0.005     1   37  15.8  -1.396e+01   1.239e+02
 6    1455    21727  0.010     5   28  14.9  -2.042e+00   1.434e+02
 7     638     8152  0.020     3   24  12.8  -4.942e-01   1.091e+01
 8     232     2074  0.039     1   17   8.9  -1.493e-04   1.138e+01
 9      66      346  0.079     1   10   5.2   8.104e-04   1.924e+01
10      10       18  0.180     1    3   1.8   2.141e-03   2.807e-03

Interpolation Matrix Information:
                 entries/row    min     max         row sums
lev  rows cols    min max     weight   weight     min       max
=================================================================
 0 303447 x 102984   0   4  -5.122e+16 3.816e+17 -5.122e+16 3.816e+17
 1 102984 x 50709   0   3   3.916e-02 1.189e+01 0.000e+00 1.189e+01
 2 50709 x 22446   0   4   2.540e-02 2.863e+00 0.000e+00 2.863e+00
 3 22446 x 8132    0   4   3.647e-02 1.670e+00 0.000e+00 1.670e+00
 4  8132 x 3055    0   4   5.507e-03 7.127e+00 0.000e+00 7.127e+00
 5  3055 x 1455    0   4   3.193e-02 1.447e+00 0.000e+00 1.447e+00
 6  1455 x 638     0   4   8.215e-03 1.461e+00 0.000e+00 1.461e+00
 7   638 x 232     0   3   1.453e-02 8.886e-01 0.000e+00 1.000e+00
 8   232 x 66      0   2   3.139e-02 9.989e-01 0.000e+00 1.000e+00
 9    66 x 10      0   2   7.679e-02 4.703e-01 0.000e+00 1.000e+00


     Complexity:    grid = 1.625239
                operator = 1.981007
                memory = 2.177350


BoomerAMG SOLVER PARAMETERS:

  Maximum number of cycles:         2
  Stopping Tolerance:               0.000000e+00
  Cycle type (1 = V, 2 = W, etc.):  1

  Relaxation Parameters:
   Visiting Grid:                     down   up  coarse
            Number of sweeps:            3    3     1
   Type 0=Jac, 3=hGS, 6=hSGS, 9=GE:      3    3     3
   Point types, partial sweeps (1=C, -1=F):
                  Pre-CG relaxation (down):   0   0   0
                   Post-CG relaxation (up):   0   0   0
                             Coarsest grid:   0

 Output flag (print_level): 3
                                            relative
               residual        factor       residual
               --------        ------       --------
    Initial    3.825253e-01                 1.000000e+00
    Cycle  1   4.985753e-01    1.303379     1.303379e+00
    Cycle  2   1.653528e+00    3.316507     4.322664e+00


 Average Convergence Factor = 2.079102

     Complexity:    grid = 1.625239
                operator = 1.981007
                   cycle = 11.886015
```


And now the same problem, same timestep, at the point that the two strategies diverge.  This time with the correct coarsening faces independently from cells:

```
BoomerAMG SETUP PARAMETERS:

 Max levels = 25
 Num levels = 5

 Strength Threshold = 0.850000
 Interpolation Truncation Factor = 0.000000
 Maximum Row Sum Threshold for Dependency Weakening = 0.500000

 Coarsening Type = Falgout-CLJP
 measures are determined locally
 No global partition option chosen.
 Interpolation = extended+i interpolation
Operator Matrix Information:

            nonzero         entries per row        row sums
lev   rows  entries  sparse  min  max   avg       min         max
===================================================================
 0  303447  3276583  0.000     6   11  10.8  -2.116e+01   1.938e+01
 1   17057   104519  0.000     1    9   6.1  -1.438e+01   1.763e+01
 2      58      154  0.046     1    6   2.7  -1.665e+01   2.199e+01
 3      17       37  0.128     1    3   2.2  -3.655e-01   1.494e+01
 4       3        5  0.556     1    2   1.7   1.764e+00   4.904e+00

Interpolation Matrix Information:
                 entries/row    min     max         row sums
lev  rows cols    min max     weight   weight     min       max
=================================================================
 0 303447 x 17057   0   2   5.530e-02 2.766e+00 0.000e+00 2.766e+00
 1 17057 x 58      0   2   1.539e-01 1.909e+00 0.000e+00 1.909e+00
 2    58 x 17      0   2   2.315e-01 1.384e+00 0.000e+00 1.384e+00
 3    17 x 3       0   1   8.398e-01 9.841e-01 0.000e+00 1.000e+00

     Complexity:    grid = 1.056468
                operator = 1.031959
                memory = 1.042412

BoomerAMG SOLVER PARAMETERS:

  Maximum number of cycles:         2
  Stopping Tolerance:               0.000000e+00
  Cycle type (1 = V, 2 = W, etc.):  1

  Relaxation Parameters:
   Visiting Grid:                     down   up  coarse
            Number of sweeps:            3    3     1
   Type 0=Jac, 3=hGS, 6=hSGS, 9=GE:      3    3     9
   Point types, partial sweeps (1=C, -1=F):
                  Pre-CG relaxation (down):   0   0   0
                   Post-CG relaxation (up):   0   0   0
                             Coarsest grid:   0

 Output flag (print_level): 3
                                            relative
               residual        factor       residual
               --------        ------       --------
    Initial    5.099872e+01                 1.000000e+00
    Cycle  1   5.756362e+00    0.112873     1.128727e-01
    Cycle  2   3.595121e+00    0.624547     7.049434e-02
```

ecoon:
As I understand it, Ifpack2 is TPetra based, not Epetra based.  And there is no Ifpack2 Hypre support.

ecoon:
Ahh, I have found a magical Ifpack2_Hypre.  It is not in Trilinos proper, but an xSDK product?

ecoon:
The Ifpack2_Hypre is nearly an identical copy of the Ifpack_Hypre implementation, and has the same bugs and consistency issues.  The last person to touch it was Alicia in 2016 -- is there a new xSDK maintainer/person?  I'll put in a pull request for Trilinos to fix Ifpack, and can add an issue/ticket for someone to update the xSDK Ifpack2 implementation if there is someone.

jd-moulton:
Wow that's a big difference.   I wonder what the asymptotic rate is?  

We should probably check in with the Trilinos team and see what is going on with interfaces to solvers.   I'm not sure who their lead is but I can ask Jim.

jd-moulton:
Ok, Alicia didn't implement your fixes correctly, and although she did understand it in the end, she left before she had a chance to do anything.

When I mentioned it to Jonathan, they said they wanted a test with it, so if you have test it would be great to give them a PR.   I'll find out who it should really be assigned to.

lipnikov:
Results sounds promising. Is original system SPD or indefinite? Either way, I would MERGE the original PR as is. I have a few problems to test block-indices.

Regarding other issues. I agree that InitializePC should do only symbolic assembly of matrices. How to achieve this faster (via Ifpack2 or re-implementation) I do not know.  Ifpack2 may be the too young package.

ecoon:
The original system is wierd.  It's certainly not SPD.  Effectively it is mfd: default Richards equation with 1st order upwinded rel perms (nonsymmetric) and an accumulation term/positive shift on the cell diagonals, and the Jacobian approximation turned on.  Then the surface equations (two-point flux on another parabolic equation) are added to the faces (of the subsurface mesh) which coincide with the surface cells.

Note that this is not an issue for "nice" meshes, but fairly easy to replicate on DEM-based (even small DEM based) triangular prism meshes.  Using the Schur complement (face-only) system doesn't have this issue, but results in a significantly smaller timestep (no Jacobian approximation term) than this approach.  I can send you a mesh if you'd be interested in trying it with Amanzi's Richards-only to see if there are issues there as well.

ecoon:
Merging this then.  It isn't efficient, but it is at least correct at this point.

PR #197, opened 2018-05-25
New interface to walkabout

lipnikov:
In the new interface, velocities and all material data are collocated at mesh vertices. I also removed all duplicated walkabout code from Flow PK.


jd-moulton:
I think its good to push this ahead on the master branch.  

But, ideally we should make these changes compatible with amanzi-0.87.    The reason is that amanzi-0.87 is the branch we are doing QA testing on and the one that we are likely to put forward at the end of the summer as a release.   It will need this capability.   If there are inherent incompatibilities making this cherry-pick over to the branch (amanzi-0.87) difficult, then we need to consider making a release of amanzi-0.88 from master this summer (maybe we should do this anyway).

lipnikov:
cherry-picking should work on amanzi-0.87.

PR #198, opened 2018-05-28
Extended BCs

lipnikov:
This PR cleans boundary conditions (BCs) in multiple places. The essential changes are as follows: 

I. Uniform approach to BCs for all operators. The new syntax is SetBCs(bc1, bc2) and ApplyBCs(primary, eliminate, essential_eqn). The meaning of  first two flags did not change, only scope of the first one (primary) has been reduced. New flag essential_eqn represents the optional math trick that returns the simplified equation x=a to the system of discrete equations in the case of essential BCs. This was befoer in the scope of "primary". In all cases currently used in the codes, primary=essential_eqn will work as before. The new call ApplyBCs(true, true, false) allows us to add flux-type BCs to off-diagonal operators in a tree-operator structure.
 
II. New BCs options have been added: TOTAL_FLUX, DIRICHLET_TYPE2, and REMOVE.
TOTAL_FLUX  is self-explaining. Example is BC -grad(p).n + p v.n = g for the advection-diffusion equation. Total flux is supported for the moment by AdvectionUpwind and PDE_DiffusionMFD (mixed formulation). Extension to other diffusion classes will be provided later as needed.
DIRICHLET_TYPE2 is needed when essential BCs are imposed in a weak sense. It is explained clearly in the header of BCs.hh. Example is the primary (without integration by parts) weak formulation of an advection problem written in a non-conservative form.
REMOVE is related to the previous option. In general, this option could be used to remove boundary contributions from a PDE for some reasons. Currently it is used by PDE_AdvectionRiemann.

III. BCs have been removed from classes Operators_*. This impacts only Operator_CellFaceSff. BCs are typically imposed before calling AssembleMatrix. If this is tre case, then Schur complement, hard-coded inside a matrix assembly, does not require second application of essential BCs. We used this class for preconditioners only.


ecoon:
Conceptually this makes sense to me -- I've been doing this in PKs, as I'm sure you have as well, but it makes sense to move it to Operators as another flag/option.  I'll need another day or so to look at the implementation -- is this time critical for you?  If so I can push off some things today, but otherwise I'll get to it tomorrow or Friday.

lipnikov:
Total flux is processed by only one operator as the Neumann-type condition. The other operators remove all boundary contributions. The operator that process flux data should be marked as leading_op=true. If you mark advection as the leading operator, ApplyBC throws assert. Note that assert is necessary, since advection should not be the leading operator in the bundle diffusion + advection.

ecoon:
I think I disagree about what the "primary" flag meant.  I did not think of "primary" as "on diagonal", but instead as meaning what you're calling "leading_op".  So for instance, in ATS's energy PK, the advection term is on-diagonal but gets primary=False.  

Under the current model, ATS does the following for advection-diffusion Neumann conditions:

1. prescribed total flux:  give Diffusion the NEUMANN flag and total flux data, primary=True. Give Advection the NEUMANN flag and 0 data, primary=False.

2. prescribed diffusive flux, advected flux is "whatever comes out": give Diffusion the NEUMANN flag and diffusive flux data and primary=True, give Advection the DIRICHLET flag and primary=False.

Under your three-flag system, is there a case where primary=True, leading=False is a different branch of the code from primary=False, leading=False?  So does primary give extra information that is independent of leading and eliminate? 


Next, on TOTAL_FLUX as a flag, is there an advection-diffusion condition that can't be expressed by some combination of NEUMANN and DIRICHLET?  I guess I don't understand line starting https://github.com/amanzi/amanzi/blob/konstantin/issue18/src/operators/PDE_AdvectionUpwind.cc#L211 -- what BC are you looking to enable with NEUMANN and !leading?

lipnikov:
Q2: "Is there an advection-diffusion condition that can't be expressed by some combination of NEUMANN and DIRICHLET? Please elaborate. This PDE requires only one condition at a point.

lipnikov:
Q1. If advection-diffusion is an off-diagonal operator, then primary must be false for both. Which one is the leading one then? We need the second flag to separate sub-operators.

ecoon:
Q1: If both are off diagonal, why do you need a leading one? Surely you don't use the off-diagonal boundary conditions in this row's equation?

As it stands, the primary usage in your branch is incorrect for the case of two diffusion on-diagonal operators.  If primary means "on diagonal", then both diffusion operators get primary=True, both set their entry to 1, both set the rhs to the value (there is only one rhs on the operator), and the discrete row equation sums the two Ops, getting 2*x = value.  Instead, if primary means what you're calling "leading", then only the "leading" diffusion op gets primary=True, and the discrete equation is 1*x=value (this is how ATS did gas diffusion with two diffusion operators on-diagonal).  Again, primary doesn't mean on-diagonal, it means "this is the operator that gets BCs applied to, be it essential BCs for try functions or fluxes for total flux conditions."

Q2: In the diffusive flux BC with inward flux, that isn't the correct thing to do. (Line 211).  A diffusive flux condition with an inward advection is a bit wierd mathematically but physically possible.  One common case of this is upwelling from depth -- the energy diffusive flux condition is the geothermal gradient, but upwelling water brings water assumed to be at equilibrium with the soil around it.  Another common case is in surface-subsurface coupling, where the diffusive flux is given by an energy balance (radiative fluxes) but the advective flux is given by a Neumann flux of water in the flow equation that comes in at a completely different temperature from the surface skin temperature (i.e. the rain comes in at air temperature, not surface skin temperature).

I'm not sure why you would negate the matrix instead of doing what is currently done in ATS, which is to give the Advective operator Dirichlet data, setting the value to the current iterate of the temperature at the boundary (in the first case) or the value of the rainfall (in the second).  So you would not pass a NEUMANN flag to the Advective operator in the case of a diffusive flux boundary condition.

lipnikov:
3 FLAGs. Consider a 2x2 tree operator where each block is D+D. You think that changing BCs for each operator, you may set up the problem correctly using only 1 flag (primary). I tend to agree with you. So potentially, you need 16 BCs for 8 operators, probably half of them is not needed. Still too many than 2 BCs which is the minimal number required from a mathematical viewpoint. Now, there is no info regarding Neumann data for off-diagonal D. The second flag (leading_op) allows you to  achieve this mathematical goal.

ecoon:
No, I'm suggesting there are bugs all over the pull request if primary implies diagonal.

Consider 2 diffusion operators on the diagonal.  In your 3-flag model, both get primary=True, as both are on the diagonal.  One gets leading=True, the other gets leading=False.  Correct?  

The code here: https://github.com/amanzi/amanzi/blob/konstantin/issue18/src/operators/PDE_DiffusionMFD.cc#L675
applies the Neumann flux anywhere primary = True, double counting the flux.  

The code at line 670 of the same file double counts the Dirichlet data.  

Both of these will be incorrect if called with primary=True for both operators because both are on the diagonal.

I believe that if you corrected all instances of primary, you would find they all changed to leading and there would be none left, eliminating the need for the "primary" flag entirely.

ecoon:
Please work through two thought examples for me.  1x1 TreeOperators, so all are on diagonal. Not a system of equations but a single PDE. 

1. DiffusionA + DiffusionB.  What are the arguments to ApplyBCs for each operator, in the case of both Neumann and Dirichlet BCs?

2. Diffusion + Advection.  Again, what are the arguments to ApplyBCs provided to these operators in the case of Dirichlet, Neumann (total flux) and Neumann (diffusive flux)?

ecoon:
Under the current codebase (master branch) they are:

```
1(Dirichlet): 
  - DiffusionA gets primary=True, eliminate=True, markers=DIRICHLET, values=Dirichlet data
  - DiffusionB gets primary=False, eliminate=True, markers=DIRICHLET, values=Dirichlet data
1(Neumann):
  - DiffusionA gets primary=True, eliminate=True (doesn't matter), markers=NEUMANN, values=data
  - DiffusionB gets primary=False, eliminate=True (doesn't matter), markers=NEUMANN, values=data
```
Note in this case primary=False of B means that, in both cases, boundary data gets applied once.

```
2(Dirichlet):
  - Diffusion gets primary=True, eliminate=True, markers=DIRICHLET, values=data
  - Advection gets primary=False, markers=DIRICHLET, values=data
2(TOTAL FLUX):
  - Diffusion gets primary=True, eliminate=True (doesn't matter) markers=NEUMANN, values=data
  - Advection gets primary=False, markers=NEUMANN, values=data
2(DIFFUSIVE_FLUX):
  - Diffusion gets primary=True, eliminate=True(doesn't matter) markers=NEUMANN, values=data
  - Advection gets primary=False, markers=DIRICHLET, values=enthalpy/concentration of boundary
```

I agree, this requires, in general, two different BC instances.  If there is a way to only ever pass the same, singular BC instance, I'm all in favor of this pull request.  I don't believe the current pull request correctly handles several of these cases.  I don't believe that it is possible to correctly handle the prescribed diffusive flux with inward advection case with only one BC instance.

lipnikov:
Q: "I don't believe that it is possible to correctly handle the prescribed diffusive flux with inward advection case with only one BC instance."  
A:  If boundary concentration in known, use it instead of diffusion flux. 1 BC. If diffusive flux is know, let the solver calculate the advective flux.  Again 1 BC. The weak formulation shows that advective matrix must be multiplied by -1 on inflow boundary. New unit test verified that the scheme was convergent for ALL boundary conditions with one BC instance.

For both D+D and D+A arguments should be as follows:
op1: primary=true, eliminate=true, leading_op=true
op2: primary=true, eliminate=true, leading_op=false
Only one BC instance must be created. I agree that double counting of BCs should be fixed. I'll modify this pull request.

ecoon:
Good, we're making progress.  Fix the bugs, and I believe you'll find that primary goes away entirely and is replaced by leading everywhere.  

On advection-diffusion, you're doing math, not physics.  I understand that when you write a parabolic PDE you need one BC.  If you've checked it, I trust you that the -1 is correctly dealing with the case, like chemical specie transport, where it never makes sense to have two BCs.

But your math is not capturing energy transport physics.  In an advection diffusion equation for energy in a coupled energy + Darcy flow, the diffusive flux is a mixture model across both solid and liquid interfaces, while the advection is in liquid phase only.  Typically the solid phase is a much larger fraction of the area, with much larger thermal mass, than the liquid phase.  So it is very possible for the solid to be a different temperature/boundary condition from the injected water -- this is used frequently in problems that Amanzi (should) care about solving.  People (and not just me, look at PFLOTRAN or FEHM or any other code that does energy+flow) allow for specification of both the diffusive flux AND temperature of the advected water to approximate the mixture model, where nearly always the fluid and liquid are assumed in equilibrium but not necessarily at the boundary, as they physically may be coming from different phases.

Mathematically, this is equivalent to reducing two superimposed equations -- one diffusion equation (energy transport in the solid phase) and one advection-diffusion equation (energy in the liquid) into a single advection-diffusion equation (energy on the continuum mixture) by assuming equilibrium.  To do this, you need two BCs, one from each equation.

I don't believe your code precludes this.  ATS can still pass the advection operator a different BC.  Just please don't break this "feature."

Also don't forget about the case where the advected quantity is not the primary variable (e.g. enthalpy, not temperature).  All cases of that need two "boundary conditions" even if one of them is a function of the other (i.e. enthalpy on the boundary as a function of temperature).  I don't think your changes break that either though.

lipnikov:
Still we need to agree on meaning, names and order of 3 flags . Any suggestions?

lipnikov:
Looks like specifying diffusicd flux and adverted water temperature, you effectively specify total flux. So I see no contradictory between math and physics.

ecoon:
I would prefer "on_diagonal" or "diagonal" to "primary" if that is your intent.

"Leading" is fine by me -- I'm not thrilled with it, but I don't have a better idea (other than "primary!")

"eliminate" makes sense to me as is, and hasn't changed in meaning.

I would prefer that all operators take all flags, even if they don't use them, to save our sanity.  Is there an obvious one to be optional?  One advantage of making none of them optional would be the code would not compile without being fixed (I prefer the compiler to find my bugs for me).  Other than that I don't much care about order.  

That said I don't think I ever use eliminate=False, though clearly it means something and could potentially be useful.

lipnikov:
will be re-open soon

lipnikov:
Modified the design of BCs to address the above critique. Description of this PR will be modified in a minute or two...

ecoon:
So now primary means "on-diagonal"?  Or it means "one special operator in this block" (AKA "leading")?  Or it means "on-diagonal + the special operator"?

In a coupled diffusion 2x2 system, does the off diagonal get "primary=True", "essential=False"?  Or "primary=False", "essential=False"?

ecoon:
Where is the right place to document these types of "conventions"?  Since they are common to both PDE_Diffusion* and PDE_Advection*, maybe the right answer is in both PDE_Diffusion.hh and PDE_Advection.hh? 

Should ApplyBCs() be made a part of the virtual interface for PDE_Advection.hh?

lipnikov:
primary has essentially the same meaning which follows closely the math:   primary=true indicates that the operator updates both matrix and right-hand side using BC data. If primary=false, only matrix is changed. 

"primary" can be used for both on and off-diagonal blocks.
For a coupled diffusion 2x2 system, the off-diagonal diffusion may use ApplyBCs(false, any, false) or
ApplyBCs(true, any, false) depending on what you want to do with fluxes of the non-primary variable.

PR #201, opened 2018-06-01
removes old ifpack patches, replaces with the new version

ecoon:
Getting CI to check this patch for me, but this is the planned Ifpack_Hypre to be submitted to Trilinos.


ecoon:
David, any progress on figuring out why my pull requests all fail on CI?

jd-moulton:
The merge you just made on PR #203 is building on Travis CI now.   I can't seem to find configuration options on how to manage the queue of CI builds.   But it does look like it is just dropping any new requests while it finishes the one its currently on.   You could wait until the current build/test finishes, then push something to a pending PR and it should build/test it.

ecoon:
But that one has already failed, for the same reasons.  Are there any CI builds going currently?  If I restart this check can we ensure nothing else is running?

jd-moulton:
It looks like it passed to me, 

[https://travis-ci.org/amanzi/amanzi/builds](https://travis-ci.org/amanzi/amanzi/builds)

what build are you seeing failed.

Issue #199, opened 2018-05-29
superlu not picked up in FindAmanzi


ecoon:
commit: 7f17b29bfbe2b9718baa04d3a11e98714a3404f7
Introduced a bug in that find_package(amanzi) no longer picks up the necessary libraries to link Amanzi libraries, namely it does not pick up the necessary superlu and superlu_dist libaries in the dependencies.

ATS uses find_package(amanzi) read in AmanziConfig.cmake, which is installed by Amanzi.  In that file, through some black magic in AmanziTargets and Amanzi_ENABLED_TPLS, Ifpack libraries are included in link_libraries, which then recursively through more black magic links in Hypre libraries, but NOT SuperLU libraries.

If Amanzi is adding SuperLU as a dependency to HYPRE, and since HYPRE doesn't provide a FindHypre to set up that dependency, Amanzi needs to make sure SuperLU is included as a linked library.

Hopefully something in there made sense to you, because it didn't to me.

ecoon:
This is not a problem on my laptop, so don't think too hard on this yet... possibly just need to better nuke TPL/Amanzi/ATS installations and try again fresh.

jd-moulton:
I'm not sure why you need to clear PR #195 to test? 

 It would be good to track down whether or not PETSc starts this bad ordering or if it is coming from somewhere  else.   I should have some time tomorrow to look at some builds.

lipnikov:
Confirmed with MPICH-3, gcc-7, and static libs. Simple solution is to move amanzi_solvers libraries to the end of the list of Amanzi libraries. 

Possible explanation. libpetsc.a was added to the list of link libraries by CMake. Since this imported library does not have dependencies, no other libraries were added.

ecoon:
Where did you reorder the amanzi libraries?  ATS's src/executables/CMakeLists.txt?  Or somewhere else?

lipnikov:
I reordered the list of amanzi libraries in ATS' src/executables/CMakeLists.txt

ecoon:
Yeah, I tried that first, and it doesn't change the order for me.  I'm a bit surprised to does for you, since all of amesos, superlu, and hypre appear in amanzi_solvers and in nothing else.  Though I'm not using PETSc in these builds, which would use superlu and hypre but not amesos and therefore might change the order.

lipnikov:
[...]

PR #202, opened 2018-06-01
Konstantin/hypre

lipnikov:
Issue #199. Added FindHypre with dependencies and replaced deprecated CMake commands. The PR  shows that CMake resolves correctly HYPRE dependencies. ATS build passed without PETSc.


ecoon:
That's great, I started going down this road of imported interface properties, but not too far.  Thanks for doing this!

ecoon:
Does this get picked up by Ifpack as well?  In the Trilinos cmake IfpackConfig.cmake it looks like the Hypre library is hard-coded -- does the Ifpack_TPL_LIST including HYPRE as a TPL mean that that dependency gets set, and therefore (since there is now a HYPRE target to resolve, and it depends upon SuperLU) this will get the full chain of dependencies?

Also, is e.g. line 163: https://github.com/amanzi/amanzi/pull/202/commits/30a51c2271647c4a5403c898fa6899107fe0fda6#diff-e6336d86377c2785203faa5cdc272e2bR163

relying on the HYPRE_DIR as the "hint" to find SuperLU correct?  I guess if you manually install HYPRE yourself, and it depends upon SuperLU, that's as good a place as any, though maybe there should be a SUPERLU_DIR hint as well in case it exists and is set by the user?

jd-moulton:
Having trouble on Wolf, will give this a try there and let you know after lunch.

lipnikov:
My understanding of CMake is that secondary dependencies for a target are processed correctly.

lipnikov:
There is no global variable SUPERLU_DIR currently. We may add it to SuperBuild/templates/amanzi-tpl-config.cmake.in or we can call amanzi with -D SuperLU_LIBRARY=<name>.

PR #203, opened 2018-06-01
adds bootstrap options ...

ecoon:
...for building RelWithDebInfo build type and for building Amanzi without all PKs to allow use by ATS users

David, just wanted you to know this existed, and make sure I didn't screw it up. "--disable-physics" now allows an Amanzi build that links to ATS.


lipnikov:
CMake variables are case-sensitive. I think we use ENABLE_Physics.

ecoon:
Wow, thanks for catching that.  The nice thing about building --with-shared I guess... there aren't issues in linking ATS builds to Amanzi (all code) shared builds in the same way there are with static builds.

jd-moulton:
This is interesting.  I assume this is used in the coupling of surface/subsurface flow.   Do we need to add a test or augment existing test.  It seems vulnerable to being broken again.

ecoon:
Ugh, I'm sorry, git fail.  My default branch wasn't on master, so a few bug fixes have been moving into this pull request.

ecoon:
ATS uses rescale capabilities for two things: 

One has to do change of variables to ensure that GMRES and other linear solvers have dot products that don't completely ignore one block.  For instance in a coupled energy-flow problem, ensuring "even" scaling across the pressure (in MPa) and temperature (in K) unknowns is important -- otherwise you have multiple orders of magnitude difference in "partial sums" to the dot products and search directions effectively are not constrained by energy error.

Two, we use rescale to deal with the fact that the surface flow diffusion operator acts on head, not pressure.  To add this operator into the subsurface (pressure variable) preconditioner, we have to rescale by dh/dp.

Yes, this should probably get a test.

lipnikov:
As I mentioned before, scaling could be a part of an iterative solver as the diagonal matrix D. It will modify the norm in the vector space Rn. In this approach, GMRES will have the 3-matrix interface. I am not sure about nonlinear solvers, but nothing prevents us from using matrix D to rescale nonlinear residual.

PR #204, opened 2018-06-05
Recent boost 

lipnikov:
This upgrades Bosst to 1.67.0 and reduces the number of boost calls in favor of std calls.


ecoon:
Why are you changing the default VerboseObject to write on rank -1?  Doesn't that, by default, silence all output through the VerboseObject?  Or am I missing a place where it defaults back to rank 0?

lipnikov:
Since the default output is on rank 0, there is no need to print "0: " in the left column. This patch should print rank only when this is explicitly requested in the input plist. Did I miss something?

jd-moulton:
Ethan, does Konstantin's response address your question?  Do you see the behavior as advertised?

ecoon:
Sorry this got lost.  Ahh, I see.  ATS doesn't have this issue -- all PKs use the alternate constructor that doesn't take a communicator, and therefore don't print the rank by default.  In ATS, only the "Debugger" object uses that constructor (as it is the only one that writes by rank, if the user asks for debug info on an non-zero rank process.

PR #207, opened 2018-06-11
Removed Set_ID and Entity_ID from GM

lipnikov:
Set_ID (= int) is used as region counter in the geometric model (GM). There is no reason to redefine int for this purpose. It refers to sets of mesh entities in mesh classes and can be potentially non-sequential.
Type Entity_ID (= int) is defined in geometry but its primary usage is in mesh classes. GM is the wrong place to define this type. 
This PR removes Set_ID and Entity_ID from GM.


lipnikov:
Yes, we have to think about long long int. Only 1 region uses GIDs as input data. First, it is not clear how to match global ids when a parallel mesh is created on a fly.  Second, these ids are read using Teuchos::Array<int>, see  RegionFactory, so they are integers now and conversion to Entity_ID is expected one way or another.

It is quite confusing when the mesh-decorated name (Entity_ID) is defined in a parallel namespace and is linked to AmanziMesh. How about using two typedefs or define one in namespace Amanzi?

PR #210, opened 2018-06-15
Added checks of min/max values to input converter

lipnikov:
Added checks of min/max values for int and double parameters in input XML file.


jd-moulton:
Please fix conflicts and we can merge this today.

PR #214, opened 2018-06-20
Modify external face mapping such that ghost faces include only bound…

dasvyat:
Currently exterior_face_map with ghosted entities includes both boundary and some internal (ghosted) faces.  I don't think it aligns with owned-ghosted relationship for other entities. In particular, ghost mapping for external_face contains entities which are not belong to "owned" mappings on other
processes.  Or in other words, the union of all ghost mapping is not equal to the union of all owned mappings. I modified external_face  ghost mapping to eliminate internal faces.

### DIFF
--- a/src/mesh/mesh_mstk/Mesh_MSTK.cc
+++ b/src/mesh/mesh_mstk/Mesh_MSTK.cc
@@ -3665,38 +3665,7 @@ void Mesh_MSTK::init_face_map()
     face_map_wo_ghosts_ = new Epetra_Map(-1,nface,face_gids,0,*epcomm_);
     extface_map_wo_ghosts_ = new Epetra_Map(-1,n_extface,extface_gids,0,*epcomm_);
 
-    //std::cout<<*extface_map_wo_ghosts_<<"\n";
-
-    // now we add ghost faces
-
-    // idx = 0;
-    // while ((ment = MSet_Next_Entry(NotOwnedFaces,&idx))) {
-    //   int gid = MEnt_GlobalID(ment);
-    //   face_gids[i++] = gid-1;
-
-    //   if (manifold_dimension() == 3) {
-    //     List_ptr fregs = MF_Regions((MFace_ptr) ment);
-    //     if (List_Num_Entries(fregs) == 1)
-    //       extface_gids[j++] = gid-1;
-    //     if (fregs)
-    //       List_Delete(fregs);
-    //   }
-    //   else if (manifold_dimension() == 2) {
-    //     List_ptr efaces = ME_Faces((MEdge_ptr) ment);
-    //     if (List_Num_Entries(efaces) == 1)
-    //       extface_gids[j++] = gid-1;
-    //     if (efaces)
-    //       List_Delete(efaces);
-    //   }
-    // }
-    // n_extface = j;
-    // nface += nnotowned;
-
-    // face_map_w_ghosts_ = new Epetra_Map(-1,nface,face_gids,0,*epcomm_);
-    // Epetra_Map *extface_map_w_ghosts_tmp = new Epetra_Map(-1,n_extface,extface_gids,0,*epcomm_);
-
-    // std::cout<<*extface_map_w_ghosts_tmp<<"\n";
-    
+   
     idx = 0;
     while ((ment = MSet_Next_Entry(NotOwnedFaces,&idx))) {
       int gid = MEnt_GlobalID(ment);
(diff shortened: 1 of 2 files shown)


lipnikov:
Only general comments since it was not I who implemented the original algorithm:
1. I suggest to remove debug code and all deprecated (original, commented-out code). Git will show us the difference anyway, but new algorithm will be more clear to read after this.
2. Is there a test that verifies directly or indirectly the proposed algorithm?

src/mesh/mesh_mstk/Mesh_MSTK.cc
```diff
@@ -3665,34 +3665,67 @@ void Mesh_MSTK::init_face_map()
...
-    n_extface = j;
-    nface += nnotowned;
 
-    face_map_w_ghosts_ = new Epetra_Map(-1,nface,face_gids,0,*epcomm_);
-    extface_map_w_ghosts_ = new Epetra_Map(-1,n_extface,extface_gids,0,*epcomm_);
+     
+    extface_map_wo_ghosts_ -> RemoteIDList(nnotowned, gl_id.data(), pr_id.data(), lc_id.data());
+
+
+    int n_extface_w_ghosts = extface_map_wo_ghosts_ -> NumMyElements();
```
raovgarimella: Is this right? The count says "with ghosts" but you are populating it with a map that says "without ghosts".

src/mesh/mesh_mstk/Mesh_MSTK.cc
```diff
@@ -3665,34 +3665,67 @@ void Mesh_MSTK::init_face_map()
     face_map_wo_ghosts_ = new Epetra_Map(-1,nface,face_gids,0,*epcomm_);
     extface_map_wo_ghosts_ = new Epetra_Map(-1,n_extface,extface_gids,0,*epcomm_);
 
-    // now we add ghost faces
-
+
```
raovgarimella: This is a useful comment to keep, I think

raovgarimella: Also, I think this code needs more comments to clarify what its doing. Its not obvious to me right away

src/mesh/mesh_mstk/Mesh_MSTK.cc
```diff
@@ -3665,34 +3665,67 @@ void Mesh_MSTK::init_face_map()
...
       int gid = MEnt_GlobalID(ment);
       face_gids[i++] = gid-1;
+    }
+    nface += nnotowned;
 
+    face_map_w_ghosts_ = new Epetra_Map(-1,nface,face_gids,0,*epcomm_);
+
+    std::vector<int> gl_id(nnotowned), pr_id(nnotowned), lc_id(nnotowned);
+   
+    idx = 0;
```
raovgarimella: Comment:
```// Build a list of ghost faces with only one cell attached - may be on exterior or processor boundary```

File: src/mesh/mesh_mstk/Mesh_MSTK.cc [hunk 4]
```diff
@@ -3665,34 +3665,67 @@ void Mesh_MSTK::init_face_map()
...
         if (efaces)
           List_Delete(efaces);
       }
     }
-    n_extface = j;
-    nface += nnotowned;
 
-    face_map_w_ghosts_ = new Epetra_Map(-1,nface,face_gids,0,*epcomm_);
-    extface_map_w_ghosts_ = new Epetra_Map(-1,n_extface,extface_gids,0,*epcomm_);
+
```
raovgarimella: Comment:
```// Get the local IDs of  (lc_id) copies of owned boundary faces on remote processors (pr_id)```

File: src/mesh/mesh_mstk/Mesh_MSTK.cc (hunk 1 above)
Comment (raovgarimella) [inline src/mesh/mesh_mstk/Mesh_MSTK.cc]: Comment:
```// Initialize ```n_ext_face_w_ghosts``` with the number of owned external faces and then update with number of ghost faces that are truly external (and not on processor boundaries)```

File: src/mesh/mesh_mstk/Mesh_MSTK.cc (hunk 4 above)
Comment (raovgarimella) [reply src/mesh/mesh_mstk/Mesh_MSTK.cc]: What does ```nnotowned``` represent here?

File: src/mesh/mesh_mstk/Mesh_MSTK.cc [hunk 5]
```diff
@@ -3665,34 +3665,67 @@ void Mesh_MSTK::init_face_map()
...
 
-    face_map_w_ghosts_ = new Epetra_Map(-1,nface,face_gids,0,*epcomm_);
-    extface_map_w_ghosts_ = new Epetra_Map(-1,n_extface,extface_gids,0,*epcomm_);
+     
+    extface_map_wo_ghosts_ -> RemoteIDList(nnotowned, gl_id.data(), pr_id.data(), lc_id.data());
+
+
+    int n_extface_w_ghosts = extface_map_wo_ghosts_ -> NumMyElements();
+       
+    for (int k=0; k < nnotowned_bnd; k++){
```
raovgarimella: This part has me confused. I think we should have a face-face chat and then add comments to say what you are doing is this section

PR #220, opened 2018-06-28
Upgraded to HDF5 1.10.1

lipnikov:
Version 1.10.2 requires CMake 3.10. David recommended to use the previous version which requires only CMake 3.2.2. Verified TPL build with clang6, gcc8, and gcc6


jd-moulton:
I think we should delay this update for a while.   It seems too soon to expect users to adopt CMake 3.10 (or above).

lipnikov:
My understanding that 1.10.1 requires CMake 3.2.2

jd-moulton:
My mistake, I wasn't reading carefully.  Is there an advantage to jumping on this new branch of hdf5 development if we need to delay moving along the branch?  Maybe we should survey NERSC and our Institutional systems to see if defaults are switching over?

lipnikov:
PR is no longer valid, HDF5 1.10.5 has been released.

PR #223, opened 2018-06-30
Added Peaceman model to converter

lipnikov:
This adds Peaceman model to spec 2.3. The model is converted to flow source "simple well" with submodel "bhp" (bottom hole pressure). Spec 2.3 introduces new  element <peaceman_well/> under liquid components of sources with standard attributes "function", "start" [s], "end" [s], "value" [Pa], and two new attributes "radius" [m] and "depth" [m]. Spec example can be also found in XML file converter_u_test3.xml


dasvyat:
I don't see any issues with this conversion.

PR #259, opened 2018-09-27
dsv/nlfv bnd

dasvyat:
New NLFV discretization which includes boundary DOF. 
Units tests for the new NLFV  verify equivalence with nlfv:default discretization.
Added capability for dynamic BC which may change type based on the switch function of time.


dasvyat:
I've removed debug information and added some comments.
I changed derivation of Operator_CellBndFace from Operator_Cell.
The main reason why PDE_DiffusionNLFVwithBndFaces is derived from PDE_Diffusion is that in spite of
all similarities  between these classes almost every function is different between these 2 classes. The similar situation with Operator_CellBndFace.cc.

lipnikov:
This PR does not compile.

Issue #286, opened 2019-03-01
Debug versions of TPLs


ecoon:
Debug versions of TPLs do not build if MOAB is built.  

Specifically MOAB does not recognize that HDF5 should exist as libhdf5_debug.a, not libhdf5.a.

ecoon:
I did a quick fix of this by linking libhdf5.a to libhdf5_debug.a.  But doing this might further confuse FindHDF5, and cause it to think we have BOTH debug and optimized builds of HDF5 available.  Doing this then (maybe only this?) caused FindHDF5 to tell us that our HDF5 libraries were:

```
HDF5_C_LIBRARIES = optimized libhdf5.a debug libhdf5_debug.a
```

In this line, "optimized" and "debug" are keywords in their library interface, e.g. 

https://stackoverflow.com/questions/5497348/debug-and-release-library-linking-with-cmake-visual-studio/5497425

Nominally this is smart -- it lets both libraries coexist, and lets cmake be smart enough to pull the right one for the right build on demand.  But it doesn't seem like these keywords can be passed in, at least not directly, to later calls, e.g. `set_target_properties(netcdf INTERFACE_LINK_LIBRARIES ...)` in FindNetCDF.  

So they have to somehow be stripped out at that point and resolved based on generater expressions (eg. https://cmake.org/cmake/help/v3.12/manual/cmake-generator-expressions.7.html )

I don't even pretend to understand cmake.

ecoon:
In the meantime, I'm punting.  I don't really want HDF5 debug build.  I really just want to be able to build Trilinos in debug mode.  So I'm going to add some options to bootstrap:

```
--debug_trilinos
--debug_tpls
--debug
``` 

And a corresponding option to SuperBuild: `TRILINOS_BUILD_TYPE` which, if defined, gets passed to Trilinos configure.  This allows, for instance:  `--debug --debug_trilinos --relwithdebuginfo_tpls` which is a Debug build of Amanzi and Trilinos but a RelWithDebInfo build of the rest of the TPLs, which (at least for me) is the right amount of debug info for debugging Amanzi infrastructure.

ecoon:
Maybe someone who understands cmake's INTERFACE stuff could fix a lot of this.  I'm not sure if our Find* are too old, too new, incompatible (we provide FindNetCDF but not FindHDF5, these should probably go together!) or what.

lipnikov:
The logic was to replace only Find* that are weak or not portable in Cmake 3.8 or lower.

Issue #284, opened 2019-02-28
clean up Mesh constructors


ecoon:
In commit bbc646113eb810ca312f9c9107ee26df447fb3a9, Konstantin did some cleanup of the mesh construction procoess.  But there is a lot more work that could be done there.

Right now we have 4 frameworks (3 of which work), with 7? constructors:

* read from file
* generate 2D
* generate 3D
* generate from GenerationSpec
* extract from reference to mesh and setnames
* extract from pointer to mesh and setnames
* extract from reference to mesh and set IDs

Logically there are really only three ways to construct, however: read, generate, or extract.

So most of these overlap and are small variations on each other.  

Furthermore, to change a constructor interface requires touching all four frameworks (which declare all constructors, but may not actually implement them), plus a "bogus_mesh" framework, plus both a create() and operator() method in the factory which just call each other, plus multiple methods in the traits implementation (four templated versions and one non-templated versions).  I believe that EACH constructor's interface definition actually appears 12 places.

So, at this point, there are three potential places to put logic about construction of a Mesh:

1. in the user's code (for instance, force the user to parse the generate parameterlist)
2. in the factory
3. in the framework's constructors

Right now we provide a rich interface in the factory (so little logic is required in user-code, good!), use the factory to choose a framework, and have many constructors.  These constructors then parse their arguments and call an internal, private Init_() method to do the actual construction.

I'm fairly sure, however, that the logic to go between, for instance, the "generate from GenerationSpec" constructor and the "generate 3D" constructor are common to all mesh frameworks.  Same for going between set names and entity ID lists.  This logic should live in one and only one place.

I believe we should be able to get down to at fewest 3, and and most 4 constructors for each mesh framework:

* read from file
* extract from Entity_ID_List

and EITHER:

* generate from GenerationSpec

OR

* generate 2D 
* generate 3D.

Then we can have our same rich interface for our factory but have many fewer constructors in the mesh.  My sense is that using the GenerationSpec is a good design idea, because it is much more extensible, but I would argue that instead of using the GenerationSpec, we should just use a ParameterList directly.  But this would also break a lot of tests, so I'm open to having 4 direct constructors in each mesh and calling one of these four from many different factory methods.

[...]

PR #287, opened 2019-03-04
Updates Amanzi to use a typedef'd Comm_type and encourages uniformity in use of Comms

### LINKED ISSUE #285: --enable-alquimia --disable-crunchtope
It appears that Alquimia requires building Crunchtope, but superbuild is not aware of that?  Or at least, running bootstrap with: `--enable alquimia --disable-crunchtope` happily builds Alquimia and doesn't build Crunchtope, but then inserts:

```
set(CRUNCHTOPE_DIR ${AMANZI_TPLS_INSTALL_DIR} CACHE PATH "CRUNCHTOPE installation prefix")
```

into amanzi-tpl-config.cmake anyway, and fails to configure Amanzi because it can't find the Crunchtop libraries.

ecoon:
This pull request references #284 and closes #285 and #286. 

All tests pass (including Simple, MSTK, and MOAB, excluding STK).

Note this touches a ton of code and slightly changes interfaces for constructing meshes.  New tests should look at how old tests now get a comm and use the mesh factory.


ecoon:
1. great, I've made that the method in all cases (generation was different previously).
2. Thank you, good catch.  Put in a unique_ptr.  Really most of the mesh frameworks should move their cache into unique_ptr() to automate via RAII
3. Also good catch (debugging cruft).  Fixed.
4. comm_p is used in the line after it is grabbed.  This could be made into a one-liner, but I think it is useful to be explicit when casts are being done, it just makes it a bit clearer what the cast is coming from.
5. yep, fixed.

ecoon:
6 is a bit trickier.  Currently those lines are set in Amanzi's TPLs config.cmake even if the TPLs were configured with ENABLE_CRUNCHTOPE=False.  

Just like every other package that we depend upon, there should be an ENABLE_* set.  I agree this needs a bit more work to make sure that SuperBuild sets those flags in the cmake. 

Yes, this will break existing TPL installations.  I'll take this out of the pull request for now, and reopen the ticket, then do a new pull request that adds this check and fixes SuperBuild to put in the required flags.

lipnikov:
6. Probably SuperBuild/CMakeLists.txt should use the same logic in XSDK and non-XSDK branches, i.e. set *_DIR only when the package is enabled

ecoon:
@jd-moulton  is this fixable?  Is there a way to force re-running the CI without a new commit?

ecoon:
I've taken out the changes to ENABLE_* and moved that discussion into #285 and will do a separate pull request there.  It will make it easier to have a small, single change if/when things break TPLs

ecoon:
Nominally the Boost stuff should have made it so that we didn't need to be as careful about the logic of what framework can construct what type of mesh from what type of file.  It should have, at compile time, detected which frameworks had which constructors.  But we don't have enough frameworks for this to really be necessary, and doing it in such a "general, compile-time traits based" way made it impossible to just read the code and understand the logic.  Oh, and it created more code than it saved.  Just a bit of over-generalization I think.

PR #293, opened 2019-03-19
Konstantin/schema

lipnikov:
Enforce strong dependence of a PDE on the underlying discretization. This affects PDEs that work with the new schema. Constructor of the new schema extracts metadata (vector of tuples) from a discretization method to create a proper CVS. The vector of metadata is also used to assemble correctly local matrices. 

Examples include DG schemes, stable schemes for Stokes flow, and various high-order conformal and non-conformal schemes.  These schemes use many DOFs associated with different geometric objects.

This is the informational PR, since you probably do not use high-order schemes. Anyway, comments are appreciated.


lipnikov:
Schema may have multiple terms even if you solve a simple diffusion equation. We have Op_Cell_FaceCell and similar names. The name defines a simple schema. However a single PDE term may need a composition of simple schemas, e.g. Cell_FaceCell + Cell_Node + Cell_CellMoments + Cell_FaceNormaComponent, etc. This information is inside new schema; hence the name "Schema". 

when you have Diffusion+Advection, both term have Operator_Schema.

lipnikov:
You are correct. There exist a simpler way to build a SuperMap related to a Schema.

PR #295, opened 2019-03-21
Upgraded Trilinos.

lipnikov:
Upgraded to version 12.14.1. Tests passed. A few compiler warnings from KokkosBlas related to C-wrapper for the function with return type std::complex<double>. We do not use complex numbers yet.


ecoon:
I'm actually using 12.15 without trouble in the Tpetra branch.  

Does 12.14 add Epetra_MultiIntVector or is that 12.15?  That would actually be useful in a few places.

lipnikov:
I do not see MultiIntVector in the source code. I think that 12.15 is not released officially.

lipnikov:
I see no interest in upgrading to Trillions 12.14. Closing tis PR.

PR #298, opened 2019-03-22
removes a lot of (dead?) code

ecoon:
This removes a ton of code that I believe is unused with the removal of STK.  At first I thought some of this got used in Simple, then I thought some of it got used in output/HDF5_MPI, but now I think it is just dead code.

This removes all of `src/mesh/mesh_data` and `src/mesh/exodus_reader`, which don't appear to be used by MOAB, MSTK, or Simple.


ecoon:
Also included in this pull request is a few minor changes to a few containers to make them follow the standard iteration notation of `begin()`, `end()`, and `size()`.

This allows, for instance,

```
for (const auto& region : *geometric_model()) {
  ...
}
```
and other shorthand for not having to write out `GeometricModel::const_iterator r = ...`

ecoon:
Fixed that and made mesh_factory do the same "copy test files into a test subdirectory" paradigm that everything else uses.  This passes a clean build for me now.

ecoon:
> Worked for me. However the purpose of exodus_reader is not obvious to me. Hope that Rao will help.

Missed this from earlier.  exodus_reader used to be used by STK.  MSTK has native readers, as does Moab.  It could be used by Simple, but that was never added, and I don't think is necessary (Simple being largely for testing only).

raovgarimella:
Yeah. Exodus_reader was a convoluted way of getting data into temporary mesh data structures before populating the real thing. Never understood why Mike chose to implement it that way.


--------------------------------------------------------------------
Rao V Garimella                              Tel: (505) 665-2928
T-5, MS B284                        '\   '\     FAX: (505) 665-5757
Los Alamos National Lab      ( )-( )  Email: rao@lanl.gov
Los Alamos, NM 87545                    Web: http://math.lanl.gov/~rao
--------------------------------------------------------------------

________________________________

PR #314, opened 2019-06-09
WIP Cache/kokkos

JulienLoiseau:
Convert cache to kokkos for gpu support


JulienLoiseau:
Cache transformed to Kokkos View/Crs (Host for now). 
Working on ARCH summit to test GPU version of the Cache.

src/data_structures/test/test_common_patterns.cc
```diff
@@ -0,0 +1,94 @@
...
+    auto wc = CreateVec("cell", Entity_kind::CELL, 1);
+
+    // compute on device
+    {
+      auto sl_view = sl_c->ViewComponent<AmanziDefaultDevice>("cell", 0, false);
+      auto poro_view = poro_c->ViewComponent<AmanziDefaultDevice>("cell", 0, false);
+      auto wc_view = wc->ViewComponent<AmanziDefaultDevice>("cell", 0, false);
+
+      typedef Kokkos::RangePolicy<AmanziDefaultDevice, LO> Policy_type;
+      Kokkos::parallel_for(Policy_type(0, sl_view.extent(0)),
```
ecoon: Will this parallel_for() be able to capture a Mesh object for use on GPUs?  I don't think so.  Need to start building some of this on Summit/Darwin/whatever to find out.

ecoon: I suspect that it may be easier to add some "bulk" accessors.

Something like:

```
// Mesh.hh in the Mesh interface:
template<class Device_type=AmanziDefaultDevice>
View<double const*, Device_type>
cell_volumes() const {
  // convert cell_volumes_ view to the right Device and return it
}
```

```
// test_common_patterns.cc
...
auto cv_view = mesh->cell_volumes<AmanziDefaultDevice>();
Kokkos::parallel_for(Policy_type(0, sl_view.extent(0)), KOKKOS_LAMBDA(const LO& i) { wc_view(i) = sl_view(i) * poro_view(i) * cv_view(i); });
```

Capturing the view in the lambda would be as expected and this provides an option for Host and Device code.

ecoon:
I merged my changes into this to keep the final merge a bit cleaner, then added a test that uses both CompositeVector and mesh to do a common "data access/operation pattern."  I'll add a few more of these as a test under `src/data_structures/test`

They currently work on CPUs, but (as noted above) I'd be a bit surprised if they worked on GPUs.

PR #321, opened 2019-07-18
Upgraded two NetCDF packages

lipnikov:
This a routine upgrade of TPLs. They fixed one bug, so one patch was reduced.


lipnikov:
If there are no comments or requests, please merge this PR.

PR #324, opened 2019-08-24
Upgraded Trilinos, added a new test for SuperLUDist

lipnikov:
Upgraded Trillions and added a test for new capabilities.


jd-moulton:
Builds of the TPLs are working on all our standard systems (OSX, Ubuntu, and NERSC).  Other build issues remain on NERSC with some Amanzi tests, but these are unrelated to this PR.  So we are almost ready on this, just one last thing ...

Reviewing the updated requirements, Trilinos requires CMake version 3.10.0 or above.  We should update our minimum requirements to be consistent.  This is in two places, the SuperBuild and Top-level CMakeLists.txt.  Then we need to verify any updated policies are behaving consistently with our builds.

lipnikov:
Global grep gives policies only in FindBoost.cmake

jd-moulton:
NetCDF update fixed build problems on Ubuntu 16.04 systems.  Everything else looks good, pushed updated build of these TPLs to dockerhub.

PR #349, opened 2019-10-08
Kokkos/function

JulienLoiseau:
- Functions working with Kokkos/GPU 
- Adding test for functions on GPU
- Correcting composite vector function test


src/functions/FunctionBilinear.cc
```diff
@@ -5,45 +5,46 @@
 namespace Amanzi {
```
ecoon: remove `#include "Epetra_SerialDenseMatrix.hh"`

ecoon: remove `#include "Epetra_SerialDenseMatrix.h"`

src/utils/HDF5Reader.hh
```diff
@@ -70,9 +73,22 @@ struct HDF5Reader {
...
+    Kokkos::resize(mat,dims[0],dims[1]); 
+    Epetra_SerialDenseMatrix m; 
+    m.Shape(dims[1],dims[0]);
+
+    // Not readable directly, need to change row major / column major 
+    //herr_t status = H5Dread(dataset, H5T_NATIVE_DOUBLE,  H5S_ALL, H5S_ALL,
+    //                        H5P_DEFAULT, mat.data());
+
     herr_t status = H5Dread(dataset, H5T_NATIVE_DOUBLE,  H5S_ALL, H5S_ALL,
-                            H5P_DEFAULT, &mat[0][0]);
```
ecoon: this should be `mat.values()` (yes, bad old code)

ecoon: Oh wait.  Sorry, make that, `Epetra_SerialDenseMatrix` --> `Teuchos::SerialDenseMatrix<double, std::size_t>` (or something like that?) and then this should read `m.values()`

ecoon:
jeez that was ugly -- I took from the title of the issue that the branch name was capitalized, not realizing it wasn't...

PR #356, opened 2019-10-24
refactors how nonlinear solvers work

ecoon:
major refactor of how nonlinear solvers work -- standardizes their interface into an abstraction where each method provides a correction, but all error control, etc, is done in a base class.  Currently tested NKA and Newton -- haven't touch JFNK, AA, Backtracking.

This brings in:

- Konstantin's changes to a norm plus a convergence monitor (i.e. can do all of L2, Linf, or user-provided ENorm for each type of convergence monitor).
- VerboseObject: LOW --> no output, MEDIUM --> output on convergence/divergence/max_iters/etc, HIGH --> output on MEDIUM plus once per iteration with the error and the L2 error.
- Standardizes input names across all solvers
- Standardizes how/when convergence tests are done, so all solvers have all options.

### DIFF
--- a/src/solvers/LinearOperatorNKA.hh
+++ b/src/solvers/LinearOperatorNKA.hh
@@ -265,8 +265,7 @@ LinearOperatorNKA<Matrix, Vector, VectorSpace>::Init(
 
   // NKA
   nka_ = Teuchos::rcp(
-    new NKA_Base<Vector, VectorSpace>(nka_dim_, nka_tol_, m_->getDomainMap()));
-  nka_->Init(plist);
+      new NKA_Base<Vector, VectorSpace>(nka_dim_, nka_tol_, m_->getDomainMap(), vo_));
 
   initialized_ = true;
 }
--- a/src/solvers/NKA_Base.hh
+++ b/src/solvers/NKA_Base.hh
@@ -74,13 +74,8 @@ namespace AmanziSolvers {
 template <class Vector, class VectorSpace>
 class NKA_Base {
  public:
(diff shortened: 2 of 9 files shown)


ecoon:
@lipnikov @dasvyat This is a pull request onto tpetra branch, not master.  It could fairly easily be backported to master, but I'd be hesitant to do so until it works for all solvers.

Conceptually, I asked for your review because I wanted to check and see if you guys were on board with this change.  Effectively it refactors nonlinear solvers to provide a virtual method to get the "update" or "du", but leaves the loop, convergence testing, etc, to a common base class, [SolverDefault.hh](https://github.com/amanzi/amanzi/blob/dev/tpetra/solvers/src/solvers/SolverDefault.hh).

I believe that all of our current solvers can fit in this framework, though if not, they could separately do all the work themselves (as is the case in our current solvers design).  This may change some input files, if inputs were not consistent across solvers, and definitely changes some defaults (e.g. Newton had default 50 max iterations, NKA had default 20.  Now the Default solver reads that parameter, so the default must be identical).

If you see any pitfalls or would like changes, letting me know now would be a big help.  Eventually we'll finish the rest of the solvers, but for now @JulienLoiseau and I are doing the minimum possible to get a reasonably working codebase.

lipnikov:
The role of the intermediate class SolverDefault is unclear, especially if all existing solvers will be derived from it. Why not to use Solver as the only base class?

All solvers are looking for good du, but the list of control parameters is different. So now they will be parsed in two places. I am okay with this.

PR #363, opened 2019-11-05
'Merge' doc from amanzi-0.87

daniellivingston:
Merge updates from the user guide on amanzi-0.87 back up to amanzi-0.88 and master branches.

All files under `doc/` on `amanzi-0.87` were effectively copied over to `amanzi-0.88`, under the assumption that all doc files under `0.87` are more up to date than `0.88`. Please review.

### DIFF
--- a/doc/developer_guide/Makefile
+++ b/doc/developer_guide/Makefile
@@ -16,7 +16,9 @@ AUXFILES=$(shell ls bu*.aux)
 $(PDFFILE): $(TEXFILES) $(BIBFILE)
 	TEXINPUTS=.//:;\
 	pdflatex $(TEXFILE);\
-	bibtex $(FILE)
+	bibtex $(FILE);\
+	pdflatex $(TEXFILE);\
+	pdflatex $(TEXFILE)
 
 $(BIBFILE): $(BIBFILES)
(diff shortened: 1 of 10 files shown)


lipnikov:
I am confused now. I do not see changes made to the user guide and located under testing/. New name for testing/ could be test_suites/

jd-moulton:
I think I mislead you about where the updates needed to come from and where the needed to move to.  All the original source *.rst files that were updated on amanzi-0.87 and need to move to their corresponding place in amanzi-0.88 are moving from the testing -> test_suite directories.  

The *.rst files in doc under user_guide are copied from testing or test_suite (depending on branch).  The supporting files under doc are likely more recent in amanzi-0.88.  We would need to look carefully at the history of these. 

Konstantin found a way to review and interactively except or decline changes through cherry-picking.   I think this would be great but not sure it would work in the case of the testing-> test_suite name change.  I think would back these changes out (or start fresh) and give this cherry-picking a try.

lipnikov:
Should we skip 0.88 branch and instead  bring master branch up-to-date?

jd-moulton:
I think amanzi-0.88 and master are very close from point of view of the User Guide.  So the end goal is to have these in sync, and I sort think it doesn't matter which goes first.   That said we seem to be pushing master closer to a new release here, so maybe it is best to do master first.

PR #378, opened 2019-12-24
Updated HDF5 package

lipnikov:
Update HDF5 to version 1.10.6


lipnikov:
Note that for the first time this upgrade reduces size of TPLs, at least on OS X

PR #379, opened 2019-12-29
Reactive transport on matrix and fractures

lipnikov:
Implemented reactive transport in matrix and fractures. Implementation supports both native and high level specs. Unit test is provided. This branch uses new interface to ASCEMIO (pending PR); hence, it requires the new stack of TPLs.


dasvyat:
Missing file ChemistryMatrixFracture_PK.hh

Issue #48, opened 2016-12-05
lifting face sets onto a surface mesh fails


ecoon:
Lifting a surface mesh from a volumetric mesh is made to be quite general in lifting labeled sets.

1. if the label is a volumetric cell label, it becomes a surface cell label (this works)
2. if the label is a volumetric face label on the side of the volume, it becomes a surface face label (this works)
3. if the label is a volumetric face label on the surface of the volume, it becomes a surface cell label.  This does not work.

Fix 3, and add this to the unit tests that does surface mesh extraction.

raovgarimella:
Look at my latest pull request #63 which actually disables inheriting a face set that the surface mesh is created from. It seems like its at odds with this ticket. In any case, before the PR it was neither doing what you consider the right thing nor what I considered the right thing.

ecoon:
To allow this to work would require updating Mesh_MSTK::build_set():
 

If you’re asking for a labeled set and this is a derived mesh (which it can know because the parent_mesh pointer is non-NULL), then:

- check the parent mesh for this label, updating the type of the entity based on manifold dimensionality (i.e. if I am dimension 2 and my parent is dimension 2+1, change cell -> face, face -> edge, etc, when querying the parent for the derived set.

- Then create a new MSet in my mesh, loop over all of my elements of that type, find their parent ID, check if it is in the MSet on the parent mesh, and stick that element in my MSet if so.

- Otherwise continue on the usual LabeledSet branch of the code.
 
This same logic would repeat a few times, at about line Mesh_MSTK.cc:2887 and Mesh_MSTK.cc:3008, and Mesh_MSTK.cc:3086.  Remember that derived meshes can be the reduced dimensionality (surface from subsurface) or also equal dimensionality (columns from subsurface, subdomains, etc).  Conceptually we intended to allow MSTK to extrude as well, which would suggest that derived meshes could be higher dimensionality, but we’ve never really fleshed that out and it would add all sorts of issues, so for now you could just deal with the cases that dimensionality was equal or one less.
 
It is a non-trivial but doable change to Mesh_MSTK.  It probably only affects one method, and requires figuring out a bit of MSTK and how to make and use MSets.  Probably a day of work to do and test for someone who knows what they are doing, but not that much longer for someone who doesn’t.

lipnikov:
[...]

PR #381, opened 2020-01-09
Lifting facesets onto surface mesh

daniellivingston:
Addresses #48. Moves critical code into `build_set` from `get_set_entities_and_vofs`.


ecoon:
Based on these tests, it looks like your changes fix set_get_entities() (yay!) but not valid_set_name().  You have some false negatives -- some of the things you say are not valid are valid.  Should be a quick fix -- I can push an updated one if you want, but you might as well finish the job since it is 90% there!

jd-moulton:
We hadn't been looking closely at valid_set_name(), we at least I haven't been.  I would say push your update to this code so we don't get bogged down here.  We have @daniellivingston starting on another case that I think is very helpful for boundary conditions on the surface mesh.  Specifically, we want a face set on the side of a 3D mesh to be lifted to an edge set on the 2D surface mesh (e.g., for outlets or non-headwater boundaries).  

Related question is if you are counting on this branch to be merged soon to master so you can use these features/fixes there.  Once you update valid_set_name(), it sounds like we could merge and pick up the other changes later.

ecoon:
@jd-moulton  and @daniellivingston , that case, i.e. a FACE of the 3D volumetric mesh going to a FACE of the 2D surface mesh is tested in my unit test and already works.

jd-moulton:
Modulo @lipnikov comment, which we can resolve in next iteration, I think this part is ready and I should accept it.  Then I think @daniellivingston can start a new branch from current master for the next phase of this work.  Let me know if you agree by updating your review status.

ecoon:
@lipnikov I'm not sure if I'd call that a false positive or not.  It returns valid_set_name() == true, but then set_get_entities() does not throw and returns an empty set.  Similarly if you ask for CELLs of a subsurface FACE set that do not coincide with the surface.  

I don't have a huge problem with returning valid_set_name() == true for these cases (since set_get_entities() doesn't error and, at least conceptually, they are valid sets), but if you think that valid_set_name() == true should instead imply a NONEMPTY set is returned from set_get_entities(), we could easily add a call to get_set_entities() in valid_set_name() and check the length.

To me the most important contract is that valid_set_name() implies that get_set_entities() won't throw.

ecoon:
@jd-moulton what is the next phase of this work?  Did you see my above comment?  I think your example already works.

jd-moulton:
We couldn't see any code that would generate an appropriate edge set in the 3D mesh and then elevate that to the derived surface mesh as a face set.  We are also pretty sure that our current ability to read labeled sets does not include edge sets.  But if some or all of these capabilities are  there that would be great.

lipnikov:
@jd-moulton  I think this code is inside inherit_labeled_set() function. It lifts edges of side sets to faces of top surface. The name implies that it will not work for other regions.

lipnikov:
@ecoon I am fine with the current behavior of valid_set_name. Regarding derived meshes, we need MeshExtractedManifold class for meshing fracture networks, due to MSTK unstable algorithms for intersecting fractures.

PR #383, opened 2020-01-14
minor updates for use of domain sets, adds override qualifiers to silence clang warnings

ecoon:
Several minor tweaks to how keys are processed:

Mostly this allows Checkpointing to happen on either a single file on MPI_COMM_WORLD or on one-file-per-process on MPI_COMM_SELF.  This is needed for column meshes, which live on MPI_COMM_SELF.  This also adds the ability to restart from such files or to grab initial conditions from such files.

### DIFF
--- a/src/mesh/mesh_mstk/MeshSurfaceCell.hh
+++ b/src/mesh/mesh_mstk/MeshSurfaceCell.hh
@@ -56,7 +56,7 @@ class MeshSurfaceCell : public Mesh {
 
   // Parent entity in the source mesh if mesh was derived from another mesh
   virtual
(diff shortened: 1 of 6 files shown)


jd-moulton:
Adding this option for writing output is great, seems like we shouldn't be using MPI_COMM_WORLD internally at all though.   But maybe I'm missing something here, and it would be another issue then anyways.  Updates on C++11 use of override seem like a good idea (I'm happy to believe Clang is a good authority here anyway).

Not sure about the perceived conflict with the src/physics/ats submodule.  I think it's because the hashes disagree between this branch and master.  If you have time could you check on this, would be good to avoid any perceived conflicts.

PR #396, opened 2020-02-26
Some modifications to independent field evaluator from file

lipnikov:
Added more error traps. Added default times.


dasvyat:
Approved

PR #432, opened 2020-05-02
Upgrade BoxLib to the next version

lipnikov:
To keep this component alive, we need to upgrade it to the last version. It will take many baby steps. This step is from 1.3.4 to 1.3.5.


jd-moulton:
Will this compile with gcc-9.x?

lipnikov:
Do not know. Why not? I can launch  a build later. The diff between
versions was 75K lines but quite innocent. We may need to make 20+ upgrades
like that

lipnikov:
The code has bugs for spacedim=3 and another set of bugs for spacedim=2. So
it may fail with a new compiler...

lipnikov:
Noticed some instability with running one structured test. This PR needs more investigation.

lipnikov:
Instability was due to the additional code added in 1.3.5 to the BoxLib initialization. It assumed silently that char** argv was not null. PR is re-opened for review.

PR #446, opened 2020-06-23
Multiscale_Transport_Alquimia

ajkhattak:


lipnikov:
Is it intended for Amanzi or ATS?

ajkhattak:
It is for ATS stream-hyporheic reactive transport. I am not sure if in Amanzi you guys prefix variable names or not (e.g., surface_0-concentration, surface_1-concentration,...), but for ATS multiscale transport model we had to prefix keys in the Alquimia_PK.cc to use it for chemistry in our subgrid model.

lipnikov:
This PR is on the Amanzi repo, not the ATS repo. It looks innocent for Amanzi's Transport except for boundary conditions and sources. Rescaling of BCs may potentially break our tests.

lipnikov:
Since I had a quick look at it, I have a short comment: state_evaluators_registration.hh andstate_evaluators_registration.hh.in are generated on a fly. No need to commit them.

ajkhattak:
You are right, other than BC and sources we have not touched Amanzi transport code. The BC and sources are modified in the TransportBoundaryFunction_Alquimia.hh and TransportSourceFunction_Alquimia.hh in order to reproduce standalone-PFLOTRAN results using ATS_Alquimia_PFLOTRAN interface. This PR is intended for Amanzi and is associated with another PR that I have submitted to ATS. If you think it's going to break something we can revisit. 
I will remove the *_registration.hh files.

lipnikov:
Also, remove .DS_Store files that were included in this PR.

Currently the PR failed to compile, see Travis CI logs. After that we need to verify that it does not alter results
of chemistry benchmarks in the user guide that are related to BCs.

lipnikov:
Also, will new BCs affect CruchFlow input data?

tools/cmake/AmanziOptions.cmake
```diff
@@ -45,7 +45,7 @@ add_feature_info(DBC
                  ENABLE_DBC
                  "Toggle design by contract (DBC) checking")
 if (ENABLE_DBC)
-  add_definitions("-D ENABLE_DBC")
+  add_definitions("-DENABLE_DBC")
```
lipnikov, line 48: New version does not work on OSX. The original version does. Any clue?

CMake Error at tools/cmake/CCSEOptions.cmake:62 (set_directory_properties):
  set_directory_properties Wrong number of arguments
Call Stack (most recent call first):
  tools/cmake/Modules/FindCCSE.cmake:38 (include)
  tools/cmake/AmanziTPL.cmake:396 (find_package)
  CMakeLists.txt:51 (include)

ecoon: "-DENABLE_DBC" works for me on mac with clang and is required by several multiple compilers.  If "-D ENABLE_DBC" works for you, this explains why I keep having to change it and keep getting grumpy cause I thought I'd fixed it already...

ecoon: Specifically nvcc rejects `-D ENABLE_DBC` (with an error message of "failed to find a file named ENABLE_DBC")

ecoon:
> This PR is on the Amanzi repo, not the ATS repo. It looks innocent for Amanzi's Transport except for boundary conditions and sources. Rescaling of BCs may potentially break our tests.

@ajkhattak Shouldn't any rescaling of BCs and sources happen in the ATS wrappers, which already apply rescalings?

tools/cmake/AmanziOptions.cmake
jd-moulton: I really that no space was the standard.  Konstantin, aside from CCSE issue, where does it error for you with the -DENABLE_DBC?

ajkhattak:
@ecoon We do have the BCs and sources wrappers but we never call them. I think if we call those wrappers then we don't need to change/modify amanzi/transport BCs and sources, and that way we can avoid breaking Amanzi use of Alquima/CrunchFlow.

src/pks/transport/TransportBoundaryFunction_Alquimia.cc
```diff
@@ -110,7 +110,7 @@ void TransportBoundaryFunction_Alquimia::Compute(double t_old, double t_new)
     // Move the concentrations into place.
     std::vector<double>& values = it->second;
     for (int i = 0; i < values.size(); i++) {
-      values[i] = alq_state_.total_mobile.data[i];
+      values[i] = alq_state_.total_mobile.data[i] / ((*mol_dens_data_)[0][cell] / 1000.);
```
ecoon, line 113: It looks like this would break Amanzi units of mol/L in favor of ATS units of mol/m^3, so it seems Ahmad will have to add an option, either in the ParameterList or in Init_() call that sets the units of these two Transport functions (Boundary and Source).

src/pks/transport/TransportSourceFunction_Alquimia.cc
```diff
@@ -86,23 +86,34 @@ void TransportSourceFunction_Alquimia::Compute(double t_old, double t_new)
...
   // Loop over sides and evaluate values.
   for (auto it = begin(); it != end(); ++it) {
     int cell = it->first; 
 
     // Dump the contents of the chemistry state into our Alquimia containers.
     chem_pk_->CopyToAlquimia(cell, alq_mat_props_, alq_state_, alq_aux_data_);
 
+    // double liquid_vol_L = alq_mat_props_.volume * (*src_liq_data_)[0][cell] * 1000 ;
+
+    double liquid_vol_L =  (*src_liq_data_)[0][cell] * 1000 ;
```
ecoon, line 100: Move this down or into the units call.

src/pks/transport/TransportSourceFunction_Alquimia.cc
```diff
@@ -86,23 +86,34 @@ void TransportSourceFunction_Alquimia::Compute(double t_old, double t_new)
...
     // Move the concentrations into place.
     std::vector<double>& values = it->second;
+    
+
     for (int i = 0; i < values.size(); i++) {
-      values[i] = alq_state_.total_mobile.data[i] / domain_volume_;
-    }
+
+      double moles = alq_state_.total_mobile.data[i] * liquid_vol_L; // conversion from mol/L to mol
+      values[i] = moles/(alq_mat_props_.volume);
```
ecoon, line 111: Here is the other place that units are broken for Amanzi.

tools/cmake/AmanziOptions.cmake
lipnikov: Ethan, last time I touched CCSEOptions.cmake was in Oct 2017 ;-). Did you build CCSE on your Mac? This seems the only package that complains.

ecoon: CCSE gets built with structured right?  I think I built it fairly recently when you said something about the bootstrap work.  Let me try again.

ecoon: Haha right, this is the exact same discussion...  it's a good thing we document this stuff now... 

https://github.com/amanzi/amanzi/pull/435#issuecomment-626118467

src/pks/transport/TransportBoundaryFunction_Alquimia.cc
lipnikov: Amanzi uses both mol/L and mol/m^3. But when Alquimia is activated, then everything should be based on geochemical conditions in PFloTran and CrunFlow input files. From this viewpoint, no scaling is needed for boundary conditions and sources. One way to proceed is to look at PFloTran files in test_suites/benchmarking/chemistry/xxx that work correctly on the master branch.

tools/cmake/AmanziOptions.cmake
lipnikov: Mistake in the CCSEOptions - incorrect list in set_directory_properties(). It is triggered only if -D has no space. Easy to fix provided that David will refresh my memory for the need of this file.

lipnikov: --- a/tools/cmake/CCSEOptions.cmake
+++ b/tools/cmake/CCSEOptions.cmake
@@ -59,7 +59,8 @@ endif(ENABLE_OpenMP)

 get_directory_property(defs COMPILE_DEFINITIONS)
 if(defs)
-  set_directory_properties(PROPERTIES COMPILE_DEFINITIONS $defs ${BL_DEFINES})
+  list(APPEND defs ${BL_DEFINES})
+  set_directory_properties(PROPERTIES COMPILE_DEFINITIONS $defs)
 endif()

src/pks/transport/TransportBoundaryFunction_Alquimia.cc
ajkhattak: Agree. My changes did break Alquimia for Amanzi. I pushed new changes that would preserve Amanzi use of Alquimia, and at the same time we will provide a key in our ATS input file when using Alquimia for ATS transport.

src/pks/transport/TransportSourceFunction_Alquimia.cc
```diff
@@ -86,23 +91,34 @@ void TransportSourceFunction_Alquimia::Compute(double t_old, double t_new)
...
 
     // Move the concentrations into place.
     std::vector<double>& values = it->second;
+    
     for (int i = 0; i < values.size(); i++) {
-      values[i] = alq_state_.total_mobile.data[i] / domain_volume_;
-    }
+
+      double moles = alq_state_.total_mobile.data[i] * liquid_vol_L; // conversion from mol/L to mol
+      values[i] = moles/(alq_mat_props_.volume);
```
lipnikov, line 116: The result does not depend on domain_volume_ unlike in the original AMANZI algorithm.

ajkhattak: my fault, I was comparing with a wrong branch. pushed a new commit to fix this issue.

lipnikov:
Please remove a few ._.DS_Store files from his commit.

ajkhattak:
@jd-moulton @lipnikov just wondering what are your plans for merging this PR?

jd-moulton:
The branch doesn't build Amanzi, see the Travis build for error messages:

<https://travis-ci.org/github/amanzi/amanzi/builds/702212046>

Does it build for you? I think we need you and @lipnikov to resolve the build errors, and ensure it's passing Amanzi tests. Then you can double check ATS regression tests and we'll check Amanzi user guide tests on  biogeochemistry.

lipnikov:
I think I mentioned the build error before. I agree with David. Since we do not have enough geochemistry tests, we need to verify user guide to re-test Amanzi with ENABLE_Regression_Tests:BOOL=TRUE

ajkhattak:
The branch builds Amanzi now, but still seeing some issues with test 139. See for example,
#1514 passed  and #1513.3 failed (test 139 - chemistry_benchmarks_1d (Failed)), however these commits are not significantly different from each other. Any thoughts? 

Screen Shot 2020-07-01 at 12 05 25 PM

Screen Shot 2020-07-01 at 12 05 39 PM

lipnikov:
Test 139 checks Ca++ volume fraction. You probably need to run this test manually (with -V option) to see the actual reason for failure.

ajkhattak:
@lipnikov @jd-moulton this PR passed all checks. Any further comments?

lipnikov:
Looks like the branch has conflicts with the master branch.

lipnikov:
This PR breaks the user guide. I looked at three chemistry benchmarks. Results of Amanzi simulations are missing. This problem is related to suffix, e.g. " conc". It is not clear why it was dropped for one field and not for the other fields (e.g. " vol frac"). With the current approach a few Python scripts in test_suite should be modified to reflect this change.

ajkhattak:
Okay, earlier I removed the suffix "conc" for the reason that a variable (or attributes) should not have any space in its name, but I reverted it to go back to the previous style.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -65,32 +65,34 @@ Alquimia_PK::Alquimia_PK(Teuchos::ParameterList& pk_tree,
...
+
+  poro_key_ = Keys::readKey(*cp_list_, domain_, "porosity", "porosity");
+  saturation_key_ = Keys::readKey(*cp_list_, domain_, "saturation liquid", "saturation_liquid");
+  fluid_den_key_ = Keys::readKey(*cp_list_, domain_, "mass density liquid", "mass_density_liquid");
+  
+
+  min_vol_frac_key_ = Keys::readKey(*cp_list_, domain_, "mineral volume fractions","mineral_volume_fractions");
+  min_ssa_key_ = Keys::readKey(*cp_list_, domain_, "mineral specific surface area","mineral_specific_surface_area");
+  sorp_sites_key_ = Keys::readKey(*cp_list_, domain_, "sorption_sites","sorption sites");
+  surf_cfsc_key_ = Keys::readKey(*cp_list_, domain_, "surface_complex_free_site_conc","surface complex free site conc");
```
lipnikov, line 78: Is it okay that default name has no underscores?

src/pks/transport/TransportSourceFunction_Alquimia.hh
```diff
@@ -39,6 +39,9 @@ class TransportSourceFunction_Alquimia : public TransportDomainFunction {
   
   void Compute(double t_old, double t_new);
 
+  void set_mol_dens_data_( Teuchos::Ptr<const Epetra_MultiVector> data) {mol_dens_data_ = data;}
```
lipnikov, line 42: I think this function is not used.

lipnikov:
> Okay, earlier I removed the suffix "conc" for the reason that a variable (or attributes) should not have any space in its name, but I reverted it to go back to the previous style.

This makes sense, but you did not do it for other suffixes, which confused me.

src/pks/chemistry/Alquimia_PK.cc
ajkhattak: Fixed this one as well.

ajkhattak:
@jd-moulton @lipnikov Seems like this PR is ready to be merged unless you guys have any more comments.

lipnikov:
If u think that Ethan's requests were resolved and there is no need to bother him again, go ahead and merge.

PR #457, opened 2020-07-05
Coupled thermal flow PK

lipnikov:
Initial implementation of the coupled thermal flow PK.


dasvyat:
Added sources to energy PKs. An additional test is required.

PR #462, opened 2020-07-14
closes blocking bug in location of Hypre and SuperLU repositories

### LINKED ISSUE #461: SuperLU moved?
This link seems to be dead:

https://github.com/amanzi/amanzi/blob/master/config/SuperBuild/TPLVersions.cmake#L412

ecoon:
The Hypre location is now dead, which makes the current master not buildable.

The SuperLU location is updated to the github site which is likely more reliable than the legacy NERSC site.

Also Build_PETSc got some new library parsing for removing spaces in list of blas libraries, required for e.g. atlas builds on Ubuntu.  This was added for lapack but not for blas, and both can have multiple libraries.

closes #461


ecoon:
Merging without review as the code is currently not buildable.

PR #463, opened 2020-07-17
Recuding cost of data copy in Alquimia

lipnikov:
This reduces run time of 2D F-Area simulation 5+ times by optimizing data copy in chemistry.


lipnikov:
Since there are no comments, I am merging.

PR #469, opened 2020-07-28
WIP: Combines Preconditioners, Direct Solvers, and Iterative Solvers

ecoon:
This rewrites the linear solvers component of the code in the following ways:

1. It creates a common interface for a base class Matrix, including:
* Apply(), DomainMap(), RangeMap()
* set_parameters(Teuchos::ParameterList&)
* UpdateInverse()  -- calling this implies that symbolic structure is set, e.g. all Ops/schemas are in place and created, all blocks are set in a TreeOperator, etc.  SymbolicAssembleMatrix() may be called.
* ComputeInverse() -- calling this implies that numeric values are now set, e.g. UpdateMatrices() has been called.  AssembleMatrix() may be called.
* ApplyInverse(Vector A, Vector B)

Things that implement this interface include: TreeOperator, Operator, test matrix classes, and all Preconditioners, DirectMethods, and IterativeMethods.   Note that not all things actually inherit from this class though -- only things that need to be stored in common pointers (e.g. PCs, DirectMethods, and IterativeMethods).

2. Splits all algorithms that supply an ApplyInverse() method into three classes (Preconditioner, DirectMethod, IterativeMethod) that are defined only by their math and not by their computer science.  All of these are templated on four types  -- the "forward operator" (called Operator) which implements Apply(), the "inverse operator" (see below), the Vector, and the VectorSpace.  

The "inverse operator" may be one of three types, depending upon context.  Which type of inverse is provided is deduced from the template argument using some enable_if calls (see src/solvers/InverseHelpers.hh).  These types can be:
  i.  "Assemblers", which provide a method with an interface that includes:  getSuperMap(), A(), AssembleMatrix(), and SymbolicAssembleMatrix(),  (e.g. TreeOperator and Operator)
  ii . "Assembled" objects which are themselves already assembled (e.g. Epetra_CrsMatrix)
  iii. "Inverse Providers" which themselves provide an implementation of ApplyInverse().  All inverse algorithms supply this (e.g. Preconditioner*, DirectMethod*, etc) but so may other helper classes (e.g. block-diagonal-preconditioned TreeOperators).

3. Update all operators and tests to use the new format.

Note, the advantages are as follows:

1. All linear solvers are gone and are included in the Operator.  So PKs no longer need to worry about whether they have a LinearOperator or a Preconditioner or both.  Just call Operator::ApplyInverse() and it will call preconditioned GMRES if that is what you constructed inside Operator.

2. All calls to SymbolicAssembleMatrix() and AssembleMatrix() are gone.  The PK cannot know if an operator _needs_ to be assembled or not.  Only the algorithm knows that (e.g. don't assemble if you're doing Diagonal preconditioned GMRES, or don't assemble if you're going to be assembled by an MPC into a block matrix).  Instead, there is an object called InverseAssembled, which calls SymbolicAssembleMatrix() and AssembleMatrix() at the right time as needed (if you called UpdateInverse() and ComputeMatrix() appropriately).  Client code should simply call InitializeInverse(), UpdateInverse(), and ComputeInverse(), and assembly will happen if needed.

All of this is templated, but largely hidden in the factory InverseFactory.  The template parameter types are deduced from the arguments passed to that factory.  Effectively there are two entry points -- one for "assemblers" and one for "inverse providers".  For TreeOperators, which may be both "assemblers" (e.g. they can assemble a global matrix) and "providers" (e.g. they provide a block diagional preconditioner themselves) I had to add a helper class so that TreeOperator::ApplyInverse(), which gets called by the MPC, can still do BoomerAMG on a global assembled matrix, or block diagonal preconditioning.

The parameter lists shouldn't have to change too much.  I did change "preconditioner type" to "preconditioning method" to match "iterative method" and "direct method".  The code errors if you do it wrong.  You also have to pass both the preconditioner and iterative method in one ParameterList now, but I added an Operator::InitializePreconditioner() variant that should work for Amanzi's use case of two separate "master" lists.

Please take a look and comment.  I haven't touched physics yet, but I did all of the operator tests to make sure our use cases are covered.  I don't think physics will be a problem (mostly it should be deleting code), but let me know if you would like changes before I get started on physics.


src/solvers/PreconditionerHypre.cc
```diff
@@ -1,46 +1,83 @@
 /*
-  Solvers
-
   Copyright 2010-201x held jointly by LANS/LANL, LBNL, and PNNL.
```
lipnikov, line 2: Since Hypre has many preconditions, BoomerAMG sounds like a more clear name.

src/solvers/IterativeMethodPCG.hh
```diff
@@ -0,0 +1,200 @@
+/*
+  Copyright 2010-201x held jointly by LANS/LANL, LBNL, and PNNL.
```
lipnikov, line 2: You git-moved GMRES but not PCG, why?

lipnikov:
Looks fine to me. Any comments on efficiency of the new approach?

src/solvers/PreconditionerHypre.cc
ecoon: We have both Boomer AMG and Euclid ILU, both of which are implemented in PreconditionerHypre (but with different options in the ParameterList).

src/solvers/IterativeMethodPCG.hh
ecoon: I did move GMRES?  Let me check and make sure the files got moved correctly in Git, I probably missed something.

ecoon:
> Looks fine to me. Any comments on efficiency of the new approach?

I don't think it changes anything efficiency wise.  The possible exception is that it improves efficiency if you called SymbolicAssemble()/Assemble() unnecessarily.  I'm not sure exactly when/if this happened, but there is a lot of code in ATS to manage calling Assemble() or not in sub-PKs that I am looking forward to getting rid of.  Effectively the MPC had to parse it's preconditioner options to find if it was going to assemble a global Jacobian, and if so, pass flags into the parameter list and down to the coupled PKs to tell them _not_ to assemble their preconditioners.

The one gotcha is if you don't refactor PKs, you'll call Assemble twice and it will work, so you won't realize you've just doubled your assembly.

Also, I ran into some issues with your Verification class in src/operators/test, so you'll see in a few places in the tests where I first initialize the PC to "Boomer AMG only", then call the CheckPreconditionerSPD() method, then re-Initialize with  PCG + BoomerAMG.  I don't think this should be an issue in production code though?  Or maybe we can figure out why the PCG + BoomerAMG would fail the SPD check (maybe just needs better tolerance?)

src/solvers/IterativeMethodPCG.hh
ecoon: Wait, are you saying that I did separate:

git rm src/solvers/LinearOperatorPCG.hh
git add src/solvers/IterativeMethodPCG.hh

instead of:

git mv src/solvers/LinearOperatorPCG.hh src/solvers/IterativeMethodPCG.hh

with the implicit issue of not tracking the history of that file?  That's possible -- let me see if I can fix that somehow?

ecoon: Ok, so I'm pretty sure I did git mv.  According to git, files are not the right things to track.  If I'd have done the move and file changes in two different commits, it would have looked a little better, but effectively I moved and edited in the same commit, and git only tracks changes.  According to git, this is a github (display/viewer) problem not a git one.

lipnikov:
> The one gotcha is if you don't refactor PKs, you'll call Assemble twice and it will work, so you won't realize you've just doubled your assembly.

What is this refectory, update of a low-level preconditioner's parameter list?
 
> Or maybe we can figure out why the PCG + BoomerAMG would fail the SPD check (maybe just needs better tolerance?)

Does the V-cycle use symmetric GS?

ecoon:
@lipnikov So I have done a lot of the physics, including all of ATS and most of Amanzi, but am getting a bit hung up on a Amanzi's flow PK.  It appears you use the same op_preconditioner_ with potentially multiple different instances of iterative methods -- you have op_pc_solver_ that gets constructed based on the parameter "preconditioner enhancement" and multiple solvers that get formed locally based off of "linear solver".  

Is this intentional and desired?  It is possible to keep this, and is potentially logical that you would want to separately define a "fully solve" option vs a "solve a little bit to weak tolerances" options, but would you ever do that in the same problem?

lipnikov:
Flow PK has complex initialization machinery: It may do saturated initialization (PCG), unsaturated initialization (GMRES), projection of a subspace of constraints (GMRES), plus enhancements.  In general, each step of the Flow PK selects an optimal solver or a solver specified by the user using different PLists.

ecoon:
> Flow PK has complex initialization machinery: It may do saturated initialization (PCG), unsaturated initialization (GMRES + Picard), projection on a subspace of constraints (GMRES), plus enhancements. In general, each step of the Flow PK selects an optimal solver or a solver specified by the user using different PLists.
> 
> Most initialization solvers are needed only for the life of the initialization object. Sorry if this design challenges your new framework. Do you have ideas to simplify things without compromising the capabilities?

In the current implementation, I believe you always share the same preconditioner, and only vary the iterative method.  Is that correct?  So you wouldn't, for instance, use AMG for a steadystate solve and then something cheaper (ILU) for a transient solve?

There are two ways to get the current behavior -- one is to use the same concept as now, where you call InitializeInverse() once, initializing *only* the preconditioner, and then instantiate (as needed) different iterative methods on top of the same preconditioner.  Effectively that is the direct refactor of the existing implementation, and this can still work.

The second would be to call InitializeInverse() multiple times, providing a PC and/or an iterative method multiple times.  Then all Inverse methods would be inside the Operator as is the case for other PKs (i.e. no second object to store in the PK, op_pc_solver_).  You could also add to Amanzi's input spect and provide different preconditioners for different stages of the solve (e.g. initialization vs transient).  The downside of course is that it would have to do all setup work both times (but likely that isn't an issue since it is only a few times).

Let me know which you would prefer.

ecoon:
On the SPD check, the issue isn't in BoomerAMG -- you were already using BoomerAMG in that check, and it passes now that way.  But instead it was failing with PCG + BoomerAMG.  Maybe a tolerance issue?  

At any rate, the unit tests currently pass using Boomer only, and then re-Initialize with Boomer+PCG for the solve part of the test.

lipnikov:
Correct, flow uses one object for the raw preconditioner, but iterative solver may vary.

Initialization of the V-cycle is costly (UpdatePreconditioner).  My understanding that both approaches do not provide savings here. If so it does, not matter much which approach will be chosen.

x^T  H  x > 0. If H is a nonlinear approximation of a A^-1, then tolerance matters.

src/operators/Operator.hh
```diff
@@ -177,14 +180,31 @@ class Operator {
...
+                         const std::string& iter_name,
+                         const Teuchos::ParameterList& iter_list,
+                         bool make_one_iteration=true);
+
+  // versions that make it easier to deal with ATS input spec format
+  void InitializeInverse(bool make_one_iteration=true);
+
+  // -- preferred methods -- three stages for init, update, and compute.
+  virtual void InitializeInverse(Teuchos::ParameterList& plist) override;
+  virtual void UpdateInverse() override;
```
lipnikov, line 205: Based solely on the name, it is not clear what is the difference between Update and Compute.

src/operators/Operator_FaceCellScc.cc
```diff
@@ -98,8 +98,7 @@ int Operator_FaceCellScc::ApplyInverse(const CompositeVector& X, CompositeVector
 
   // Solve the Schur complement system Scc * Yc = Tc.
```
lipnikov, line 99: The comment does not match the code. Please double-check.

src/operators/TreeOperator.cc
```diff
@@ -103,28 +110,38 @@ int TreeOperator::ApplyAssembled(const TreeVector& X, TreeVector& Y) const
 /* ******************************************************************
 * Calculate Y = inv(A) * X using global matrix.
 ****************************************************************** */
-int TreeOperator::ApplyInverse(const TreeVector& X, TreeVector& Y) const
+int TreeOperator::ApplyInverse(const TreeVector& Y, TreeVector& X) const
```
lipnikov, line 113: Looks unnatural to apply operator to Y and to return X.

src/operators/TreeOperator.cc
```diff
@@ -103,28 +110,38 @@ int TreeOperator::ApplyAssembled(const TreeVector& X, TreeVector& Y) const
...
+                                     // have been created.
+    return ApplyInverseBlockDiagonal_(Y,X);
   }
+}
+
 
+/* ******************************************************************
+* Calculate Y = inv(A) * X using the block diagonal
+****************************************************************** */
+int TreeOperator::ApplyInverseBlockDiagonal_(const TreeVector& Y, TreeVector& X) const
```
lipnikov, line 134: Again, the comment does not match "const" in arguments.

src/pks/flow/Darcy_PK.cc
```diff
@@ -545,13 +541,7 @@ bool Darcy_PK::AdvanceStep(double t_old, double t_new, bool reinit)
     p_old = Teuchos::rcp(new Epetra_MultiVector(*solution->ViewComponent("cell")));
   }
 
-  // create linear solver and calculate new pressure
-  AmanziSolvers::LinearOperatorFactory<Operators::Operator, CompositeVector, CompositeVectorSpace> factory;
-  Teuchos::RCP<AmanziSolvers::LinearOperator<Operators::Operator, CompositeVector, CompositeVectorSpace> >
-     solver = factory.Create(solver_name_, *linear_operator_list_, op_);
-
-  solver->add_criteria(AmanziSolvers::LIN_SOLVER_MAKE_ONE_ITERATION);
```
lipnikov, line 553: What happened to this options?

src/pks/mpc_pk/FlowMatrixFracture_PK.cc
```diff
@@ -288,22 +293,19 @@ void FlowMatrixFracture_PK::UpdatePreconditioner(double t,
...
-  op_tree_->InitPreconditioner(pc_list);
 }
 
 
 /* ******************************************************************* 
 * Application of preconditioner
 ******************************************************************* */
-int FlowMatrixFracture_PK::ApplyPreconditioner(Teuchos::RCP<const TreeVector> X, 
-                                               Teuchos::RCP<TreeVector> Y)
+int FlowMatrixFracture_PK::ApplyPreconditioner(Teuchos::RCP<const TreeVector> Y,
```
lipnikov, line 304: I do not understand why X and Y changed places. Looks weird.

src/operators/Operator.hh
ecoon: I agree, but I am following Trilinos convention here.  I'll add some better documentation and comments.

Effectively:

* `Initialize()` or `SetParameters()` or `initialize()` or `set_parameters()` in Trilinos solvers simply sets, validates, and parses the ParameterList.

* `update()` indicates that the sparsity structure is now set and/or has changed since the last call to `update()`.  Any work that can be done only based on the graph can now be done (e.g. calling a partitioner for a coloring for Gauss-Seidel, etc)

* `compute()` indicates that the numeric values are now set and/or have changed since the last call to `compute()`.  Any work based on values can now be done (e.g. factorizations).

* `apply()` or `ApplyInverse()` does the inverse.


At each level, a call to one function invalidates everything below it (e.g. changing sparsity invalidates values) but not visa versa (e.g. changing values can use the same structure) and tries to minimize replication of work.

src/operators/TreeOperator.cc
ecoon: I actually think this makes more sense... Apply(X,Y) --->  Y = A*X, ApplyInverse(Y,X) --->  X = A^-1*Y.  This keeps consistency across the two methods.  I was going through TreeOperator and looking at maps (see my below comment on the ?bug? in the coupled flow MPC), and want to be able to check that X is from the DomainMap() and Y from the RangeMap().

src/pks/mpc_pk/FlowMatrixFracture_PK.cc
```diff
@@ -143,7 +143,7 @@ void FlowMatrixFracture_PK::Initialize(const Teuchos::Ptr<State>& S)
   cvs_matrix->SetMesh(mesh_matrix)->SetGhosted(true)
             ->AddComponent("face", AmanziMesh::FACE, Teuchos::rcpFromRef(mmap), Teuchos::rcpFromRef(gmap), 1);
 
-  cvs_fracture->SetMesh(mesh_matrix)->SetGhosted(true)
+  cvs_fracture->SetMesh(mesh_fracture)->SetGhosted(true)
```
ecoon, line 146: I think this was a bug -- can you confirm @lipnikov

Were these two meshes the same?  I'm not sure how the test passed before with the wrong mesh -- it resulted in memory/valgrind issues for me.

src/pks/mpc_pk/TransportMatrixFractureImplicit_PK.cc
```diff
@@ -168,7 +167,14 @@ void TransportMatrixFractureImplicit_PK::Initialize(const Teuchos::Ptr<State>& S
   pk_matrix->op_adv()->ApplyBCs(true, true, true);
   pk_fracture->op_adv()->ApplyBCs(true, true, true);
 
-  op_tree_->SymbolicAssembleMatrix();
+  std::string name = tp_list_->sublist("time integrator").get<std::string>("preconditioner", "Hypre AMG");
+  std::string ls_name = tp_list_->sublist("time integrator").get<std::string>("linear solver", "none");
```
ecoon, line 171: Please let me know if you would like these defaults to change.  I was hesitant to put your hard-coded "GMRES with Hypre AMG" value in because the tests did not have this list in their xml file.  

What behavior does Amanzi expect if the solver/preconditioner name (e.g. "AztecOO CG", or "GMRES with Hypre AMG") does not exist in the solvers or preconditioners global list?  Currently this errors, and this was the only place where the sublist "gmres parameters" was created on the fly, which bypassed the error and provides an empty "default parameters" GMRES.

src/pks/mpc_pk/FlowMatrixFracture_PK.cc
ecoon: See above.  Happy to flip these back as they are in physics code though.  I was trying to figure out a valgrind issue in maps not matching (which I believe was because of the bug just below this).

src/operators/TreeOperator.cc
lipnikov: I thought about this when I wrote “unnatural”. Without both operations in one code block, this is overkill

src/pks/mpc_pk/FlowMatrixFracture_PK.cc
lipnikov: When we have only x and y, I would always write y = A x, regardless of meaning of A

src/pks/mpc_pk/FlowMatrixFracture_PK.cc
lipnikov: I think it is fixed on a branch.

src/solvers/test/solver_linear_operators.cc
```diff
@@ -230,36 +248,36 @@ TEST(NKA_SOLVER) {
   delete comm;
 };
 
-TEST(BELOS_GMRES_SOLVER) {
-  std::cout << "\nChecking Belos GMRES solver..." << std::endl;
+// TEST(BELOS_GMRES_SOLVER) {
```
lipnikov, line 251: This test has to be restored before merge.

src/pks/transport/TransportExplicit_PK.cc
```diff
@@ -326,23 +325,32 @@ bool TransportExplicit_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
     Teuchos::RCP<Operators::PDE_Accumulation> op2 =
         Teuchos::rcp(new Operators::PDE_Accumulation(AmanziMesh::CELL, op));
 
+    // Create the preconditioner and solver.
+    //
+    // Both the previous and current implementation re-create the Ops each
+    // timestep.  Is it necessary to "new" the PDE_Diffusion every timestep?
+    // Effectively this means that everything must be allocated and destroyed
+    // every timestep (local matrices, global matrices/vectors, Krylov vectors,
+    // preconditioner workspace).  --etc
```
lipnikov, line 343: I see no way to safe work when diffusion tensor is changing on each time step and for each species.

src/operators/Operator.hh
ecoon: Hmm, ok, after looking around some more, this isn't as "standard Trilinos" as I thought.

I think I will merge Initialize() and Update(), and go back to our two-stage approach (but with slightly different names): InitializeInverse() recreates all memory structure and sets the symbolic structure, and ComputeInverse() implies that the matrix values have changed.  This is a nod to the fact that, after going through the physics, nowhere do we not call InitializeInverse() and UpdateInverse() back to back at this point -- effectively we should just merge them.

src/operators/Operator_FaceCellScc.cc
ecoon: Yeah, I'm not totally sure what is going on there.  I didn't change the comment or the order of operands here, so the error is old.

src/pks/flow/Darcy_PK.cc
ecoon: Darcy_PK.cc line 425, the last argument (true) indicates to add this option.  Effectively I did not want to bring every interface from every PC, iterative method, direct solve, operator, etc, into the "common" interface.  You could cast the thing to an IterativeMethod if you wanted to get at that interface, but the helper function called in that line of code sets the criteria to include MAKE_ONE_ITERATION.

src/pks/mpc_pk/FlowMatrixFracture_PK.cc
ecoon: I don't agree -- Ax = y is quite standard notation, and so x = inv(A)*y makes more sense to me.  The fact that there is both Apply and ApplyInverse (and therefore A is a forward operator) is implied in the name ApplyInverse().  Whether inv() is "true inverse" or "approximate inverse" has no bearing on the question.  But it isn't that important, so since it bothers you I've changed it back.

Even worse though is our use of ApplyInverse(v, hv), which is a flat out lie since h is itself the variable name for the preconditioner (not the inverse of the preconditioner), e.g. solver factories take two arguments, "m" and "h".  So it should be ApplyInverse(v, h_inv_v) or something.  But that looks terrible...

src/pks/transport/TransportExplicit_PK.cc
ecoon: Can't you call SetTensorCoefficient() again?  I mean, SetTensorCoefficent() _should_ (but doesn't) flip the `mass_matrices_initialized_` flag.  But off the top of my head, without thinking too hard about it, I would think you could flip that flag, recreate mass matrices, and everything else would be fine (e.g. all Ops are already allocated, all maps are fine, etc).

src/solvers/test/solver_linear_operators.cc
ecoon: still need to add this

src/operators/Operator_FaceCellScc.cc
lipnikov: "suffix" c in the comment is no longer explicitly in the code, since ViewComponent("cell") is not used. I might extend the comment line later.

src/pks/flow/Darcy_PK.cc
lipnikov: I just wanted to double-check. This option is critical for Theis problem.

src/pks/transport/TransportExplicit_PK.cc
lipnikov: Yes, I could save on some maps and containers. But, since most time is spend on mass matrices and V-cycle initialization, for code clarity, the presented approach was selected. I have nothing against your approach but it is not priority now, so I'd suggest to make an issue for the future.

Issue #453, opened 2020-06-30
TreeOperator of TreeOperators


lipnikov:
The current concept of the TreeOperator is not easy to extend to a hierarchical case when a TreeOperator of TreeOperators is needed.  One problem is Apply(x, y) where x and y are some TreeVectors. For a hierarchical implementation, we need to switch to the CV and Operator interfaces, which is feasible but not so great implementation-wise.  Do you see any alternatives?

ecoon:
I'm not sure I understand what aspect of the CV and Operator interfaces you're looking for.  Is it just the double optional argument to Operator::Apply?  That seems straightforward.

I agree that this doesn't deal with being hierarchical right now because SetOperatorBlock() and its other block stuff is all Operators instead of recursively TreeOperators.  Is that the problem?

lipnikov:
From the comment section: "Note that a TreeVector may have EITHER subvecs OR data..." Thus, we need to fork the code to use different data access, and either TreeOperator::Apply(TV, TV) or Operator::Apply(CV, CV).

lipnikov:
SetOperatorBlock() could be used for an approach alternative to a hierarchical approach. But it requires tools like RCP<TV> to std::vector<RCP<CV>> conversion. I want to discuss options before writing more code.

ecoon:
> From the comment section: "Note that a TreeVector may have EITHER subvecs OR data..." Thus, we need to fork the code to use different data access, and either TreeOperator::Apply(TV, TV) or Operator::Apply(CV, CV).

There does need to be a fork, but the interface is consistent -- TreeOperator::Apply only accepts TreeVectors.  If both the domain and range are leaf nodes, then the TreeOperator has a non-null RCP<Operator>, and implements Apply by calling Operator::Apply().  If either domain or range are not leaf nodes, then it has a size(domain) x size(range) 2D array of TreeOperators, which may themselves either be leaf or internal nodes, where size() is defined by:

```
int size(TVS tvs) {
    int s = tvs.NumChildren();
    if (tvs.Data() != null) s++;
    return s;
}
```

[...]

PR #487, opened 2020-08-28
Extends TreeOperator to work hierarchically

ecoon:
see #453 

Note this also adds a new Op, which may be useful for non-square mesh-to-mesh coupling fluxes.  It currently is only used in the new test (which starts to hint at how one could couple surface and subsurface using two operators and non-square off-diagonal blocks), operators_tree.


ecoon:
merging from master, and checking regression tests.  should be ready by tomorrow unless there are review comments.

src/operators/Op_MeshInjection.hh
```diff
@@ -0,0 +1,82 @@
...
+
+class Op_MeshInjection : public Op {
+public:
+  Op_MeshInjection(const Schema& schema_row, const Teuchos::RCP<const AmanziMesh::Mesh>& mesh_row,
+            const Schema& schema_col, const Teuchos::RCP<const AmanziMesh::Mesh>& mesh_col,
+            const Teuchos::RCP<Epetra_Map>& injection_,
+            bool transpose_=false)
+    : Op(schema_row, schema_col, mesh_row),
+      mesh_col_(mesh_col),
+      injection(injection_),
```
lipnikov, line 33: typically, we use a_(a) instead of a(a_).

src/data_structures/TreeVector.cc
```diff
@@ -104,6 +104,36 @@ int TreeVector::PutScalar(double scalar) {
...
 }
 
+int TreeVector::PutScalarMasterAndGhosted(double scalar) {
+  // Set all data of this node and all child nodes to scalar.
+  int ierr = 0;
+  if (data_ != Teuchos::null) {
+    ierr = data_->PutScalarMasterAndGhosted(scalar);
+    if (ierr) return ierr;
+  }
+  for (std::vector< Teuchos::RCP<TreeVector> >::iterator subvec = subvecs_.begin();
```
lipnikov, line 114: Optional: I use "auto" in for-loops for clarity, but it is personal.

src/operators/TreeOperator.cc
```diff
@@ -27,78 +27,179 @@
...
+
+TreeOperator::TreeOperator(Teuchos::ParameterList& plist)
+  : TreeOperator()
+{
+  vo_ = Teuchos::rcp(new VerboseObject("TreeOperator", plist));
+  if (plist.isSublist("inverse")) set_inverse_parameters(plist.sublist("inverse"));
+}
+
+
+TreeOperator::TreeOperator(const Teuchos::RCP<const TreeVectorSpace>& row_map,
```
lipnikov, line 60: how about row_tvs and col_tvs to be crystal clear?

src/operators/TreeOperator.hh
```diff
@@ -108,35 +155,23 @@ class TreeOperator : public Matrix<TreeVector,TreeVectorSpace> {
...
-  
+int ApplyInverseBlockDiagonal_(const TreeVector& X, TreeVector& Y) const;
 
  private:
-  friend Impl::TreeOperator_BlockPreconditioner;
+  friend Impl::TreeOperator_BlockDiagonalPreconditioner;
+
+  Teuchos::RCP<const TreeVectorSpace> row_map_, col_map_;
+  std::size_t row_size_, col_size_;
+  Teuchos::Array<Teuchos::Array<Teuchos::RCP<TreeOperator> > > blocks_;
```
lipnikov, line 168: consistency inside a file: either "> >" or ">>"

src/operators/test/AnalyticElasticityBase.hh
```diff
@@ -55,6 +60,101 @@ class AnalyticElasticityBase {
...
+
+  for (int c = 0; c != ncells_owned; ++c) {
+    mesh_->cell_get_nodes(c, &nodes);
+    int nnodes = nodes.size();
+
+    for (int i = 0; i < nnodes; i++) {
+      vol_node[0][nodes[i]] += mesh_->cell_volume(c) / nnodes;
+    }
+  }
+  vol.GatherGhostedToMaster("node");
```
lipnikov, line 87: Is vol used?

src/operators/test/operator_stokes.cc
```diff
@@ -32,14 +32,224 @@
...
 
 #include "AnalyticElasticity02.hh"
 #include "Verification.hh"
 
+
 /* *****************************************************************
 * Stokes model: exactness test.
 ***************************************************************** */
-TEST(OPERATOR_STOKES_EXACTNESS) {
+SUITE(OPERATOR_STOKES) {
```
lipnikov, line 44: Make is ticket for this before merging PR. I'll also look at it.

src/solvers/DirectMethodAmesos.cc
```diff
@@ -135,44 +135,43 @@ void DirectMethodAmesos::ComputeInverse() {
...
   Epetra_Vector* vv = const_cast<Epetra_Vector*>(&v);
   problem_->SetRHS(vv);
   problem_->SetLHS(&hv);
-  
+
   if (returned_code_ == 0) {
     int ierr = solver_->Solve();
     if (ierr) {
-      returned_code_ = -ierr;
+      returned_code_ = ierr;
```
lipnikov, line 147: I remember that sign was changed for a reason. But now, the return value is 0 or 1, so should be fine. correct?

src/operators/Op_MeshInjection.hh
ecoon: Depends on whether the internally stored variable is public or private.  Private variables (and methods) should end in an "_" while public ones should not.

Op classes were originally structs that had no methods and no private data -- only public data for the schema, matrices, and diag.  Since then we've started to add more methods and private data, but haven't changed everything over to a more typical class, so they're halfway between classes and structs.

src/data_structures/TreeVector.cc
ecoon: Yeah, I agree, more auto for these types of loops is a good thing.  Will change it.  A lot of the loops in TreeVector would benefit from switching to the much cleaner:

```
for (const auto& thing : container) {
  ....
}
```

notation but they were pre-C++11 code.  Will fix them all now.

src/operators/test/AnalyticElasticityBase.hh
ecoon: Removed.  Hopefully you get a chance to look at that test (operators_stokes) where I tried to add a forward apply check to tests TreeOperator and couldn't get it working on master -- I am not sure why it doesn't seem to work there.  It looked like possibly the boundary conditions were the issue -- if we fill the vector in the usual way, calculate Au - rhs, then the only non-small values were on boundaries, in (if I recall correctly) the normal flux component of the vector.

Likely I just don't completely understand the system of equations being solved there, but I left it in case you wanted to take a look.

src/operators/TreeOperator.cc
ecoon: I went with map instead of tvs to be consistent across the interface.  I'd like all of our operators to have as similar an interface as possible, e.g. `get_row_map()` in both Operator and TreeOperator, as opposed to `get_row_tvs()` and `get_row_cvs()`.

So if the getters are then fixed, I think it is better to stay consistent across the variable name and the getter, e.g. if it is `get_row_map()`, then the internally stored variable should be called `row_map_` and the constructor argument should be `row_map`.

src/operators/test/operator_stokes.cc
ecoon: thanks

src/solvers/DirectMethodAmesos.cc
ecoon: Yeah, we were inconsistent across the linear operators of what we returned there.  Some (our iterative methods) used the strategy of positive values indicate convergence, 0 indicated still going, and negative indicated errors.  Others used 0 = success, 1 = failure.  Still others returned 1 on success and 0 on failure (which was leftover from a previous attempt at standardizing).

Effectively I standardized all of them to return 0 on success and 1 on failure, and store their (algorithm dependent) return code, which you can get via `returned_code()` or `returned_code_string()`.  I think this is the most natural for client code who just want a "pass/fail" flag, and consistent with the rest of our interfaces which always have `ierr = 1` implies failure.  

Since this lets us allow each algorithm to have their own idea of what the code means, there was no point in negating this one, and it was potentially confusing to do so.

ecoon:
closes #453

PR #490, opened 2020-09-02
refactor of ATS (reactive) transport

ecoon:
This is part of a much broader update of ATS's transport functionality.

ATS maintains its own Transport PK to enable the use of evaluators -- this allows ATS to do transport in streams, where for instance, porosity doesn't make much sense but ponded_depth is a quantity that is needed instead.  It also allows ATS's transport to impose an input spec that looks more like other ATS PKs.

However, all other functionality (e.g. dispersion coefficient models, DomainFunctions, and Alquimia wrappers) can be shared with Amanzi.  Furthermore, the Chemistry_PK and derivatives can be shared with Amanzi.

This refactors ATS to rip out all of the duplicates of that functionality and use Amanzi's.  Several minor changes were made to Amanzi to allow this, including prefixing a few variable names with a domain, allowing parameters to come from more than one places (as Amanzi expects many things to appear in the State list, while ATS expects them to appear in the PK list, to keep this information local to the PK and enable multiple instances, e.g. geochemical conditions that are different in the stream from in the subsurface).  

The only change to Amanzi is in the Chemistry_PK.  In that PK, molar density and porosity were already evaluators, but saturation was not.  Saturation was made to also be an evaluator.  This changes in Amanzi's native spec how saturation appears -- in several tests it was a variable with no evaluator, while now it must be an evaluator.  I'm not sure whether this affects any actual Amanzi runs.


src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -98,7 +98,11 @@ Alquimia_PK::Alquimia_PK(Teuchos::ParameterList& pk_tree,
   Teuchos::RCP<Teuchos::ParameterList> state_list = Teuchos::sublist(glist, "state", true);
 
   InitializeMinerals(cp_list_);
-  InitializeSorptionSites(cp_list_, state_list);
+  if (cp_list_->isSublist("initial conditions")) {
```
lipnikov, line 101: A fragile check. "initial conditions" list is Amanzi is the remnant of old times and may go away with a refectory of the chemistry converter. Since ATS keeps it local, I'would use an ATS-based criterion.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -423,16 +432,20 @@ void Alquimia_PK::XMLParameters()
...
     Exceptions::amanzi_throw(msg);
   }
-  Teuchos::ParameterList& state_list = glist_->sublist("state");
-  Teuchos::ParameterList& initial_conditions = state_list.sublist("initial conditions");
-  Key geochemical_list_key =  Keys::readKey(initial_conditions, domain_, "geochemical conditions","geochemical conditions");
-  if (initial_conditions.isSublist(geochemical_list_key)) {
-    Teuchos::ParameterList geochem_conditions = initial_conditions.sublist(geochemical_list_key);
+
+  Teuchos::RCP<Teuchos::ParameterList> initial_conditions;
+  if (cp_list_->isSublist("initial conditions")) {
```
lipnikov, line 437: Something is backwards compared to the previous check. Below is output of the current converter (test #4). 
I think that the converter should remove sublist "initial conditions" completely.  The test is common/interface_platform/converter_u_base.

"""" 
<ParameterList name="chemistry">
      <Parameter name="domain name" type="string" value="domain"/>
      <Parameter name="engine" type="string" value="PFloTran"/>
      <Parameter name="chemistry model" type="string" value="Alquimia"/>
      <Parameter name="engine input file" type="string" value="test/converter_u_test4.in"/>
     <ParameterList name="initial conditions">
        <ParameterList name="first_order_decay_constant">
          <Parameter name="function" type="string" value="list was moved to state"/>
        </ParameterList>
"""

src/pks/chemistry/CMakeLists.txt
```diff
@@ -110,14 +110,18 @@ if (BUILD_TESTS AND ENABLE_Unstructured)
...
     endif()
 
-    # The following test is disabled in parallel environments. It has to be rewritten in order 
-    # to function properly there.
-    add_amanzi_test(chemistry_benchmarks_1d chemistry_benchmark_1d
-                    KIND int
-                    SOURCE test/chemistry_benchmarks_1d.cc
-                    LINK_LIBS chemistry_pk state pks data_structures
+    if (ENABLE_AmanziPhysicsModule)
+      # this test uses the Amanzi executable...
```
lipnikov, line 114: Redundant, since it says above AmanziPhysics.

src/pks/chemistry/Chemistry_PK.cc
```diff
@@ -40,22 +40,18 @@ void Chemistry_PK::Setup(const Teuchos::Ptr<State>& S)
...
-      ->AddComponent("cell", AmanziMesh::CELL, 1);
-  }
-  
-  if (!S->HasField(fluid_den_key_)) {
-    // S->RequireScalar(fluid_den_key_, passwd_);
-    S->RequireField(fluid_den_key_, passwd_)->SetMesh(mesh_)->SetGhosted(false)
-       ->AddComponent("cell", AmanziMesh::CELL, 1);
-    S->RequireFieldEvaluator(fluid_den_key_);
-  }
+  S->RequireField(poro_key_, passwd_)->SetMesh(mesh_)->SetGhosted(false)
```
lipnikov, line 43: If some PK has already requested porosity field and/or evaluator, is it safe to do it again here for the second time?

src/pks/chemistry/Chemistry_PK.cc
```diff
@@ -40,22 +40,18 @@ void Chemistry_PK::Setup(const Teuchos::Ptr<State>& S)
...
+    ->AddComponent("cell", AmanziMesh::CELL, 1);
+  S->RequireFieldEvaluator(poro_key_);
+
+  S->RequireField(saturation_key_, passwd_)->SetMesh(mesh_)->SetGhosted(false)
+    ->AddComponent("cell", AmanziMesh::CELL, 1);
+  S->RequireFieldEvaluator(saturation_key_);
+
+  // S->RequireScalar(fluid_den_key_, passwd_);
```
lipnikov, line 51: Remove

src/pks/transport/MDM_Isotropic.cc
```diff
@@ -26,7 +28,40 @@ namespace Transport {
...
+    dispersivity_ = true;
+    count++;
+  }
+  if (plist.isParameter("dispersion coefficient [m^2 s^-1]")) {
+    alpha_ = plist.get<double>("dispersion coefficient [m^2 s^-1]");
+    dispersivity_ = false;
+    count++;
+  }
+
+  if (plist.isParameter("dispersion coefficient")) {
```
lipnikov, line 46: Haha... okay for a single occurrence, but is not a good strategy if we continue to mix and match.

src/pks/transport/TransportBoundaryFunction_Alquimia.cc
```diff
@@ -96,31 +91,25 @@ void TransportBoundaryFunction_Alquimia::Init_(const std::vector<std::string>& r
 /* ******************************************************************
 * Evaluate values at time.
 ****************************************************************** */
-void TransportBoundaryFunction_Alquimia::Compute(double t_old, double t_new)
```
lipnikov, line 99: Great!

src/pks/transport/TransportDomainFunction_UnitConversion.hh
```diff
@@ -0,0 +1,67 @@
+/*
+  Transport PK
+
+  Copyright 2010-201x held jointly by LANS/LANL, LBNL, and PNNL.
+  Amanzi is released under the three-clause BSD License.
+  The terms of use and "as is" disclaimer for this license are
+  provided in the top-level COPYRIGHT file.
+
+  Authors: Ethan Coon (coonet@ornl.gov)
```
lipnikov, line 9: Please add more comments about which units to which units.

src/pks/chemistry/Alquimia_PK.cc
ecoon: That would be great.  The naming of "initial conditions" in state was a bad name on my part (almost a decade ago) when it should have been something like "global constants".

ecoon: Oh wait, this is flat out an error, but our ATS testing didn't have sorption sites -- I intended it to be:

```
  if (cp_list_->isSublist("initial conditions")) {
    InitializeSorptionSites(cp_list_, cp_list_);
  } else {
    InitializeSorptionSites(cp_list_, state_list_);
  }
```

But Amanzi's test uses both "initial conditions" lists.  ATS will put all of this stuff in the PK's list.  

Hmm, I'm not sure how to robustly check this then other than to add an `#ifdef ENABLE_AmanziPhysicsModule`?

ecoon: I guess we could just check both lists for all parameters...

src/pks/chemistry/Alquimia_PK.cc
ecoon: Wait, you're saying you want Amanzi to remove PK "initial conditions" list completely?  I thought you were moving away from state's "initial conditions" list, but you're going the other direction?

Personally I prefer to keep things local.  Hence, in ATS, all solvers, operators, initial and boundary conditions -- everything needed to define the PDE are self contained within the PK's list.  This way, when you have many PKs, you can still see a complete story of one PDE by looking at one list, instead of having to look at the global xml file.  Plus it makes reuse of input files easier -- you can just copy paste a PK list and have the complete component.

I guess the right way to deal with this is going to have to be either:

1.  `#ifdef ENABLE_{ATS,Amanzi}PhysicsModule` in the source or
2.  have a parameter in the PK list e.g. `bool input_format_ats = plist.get("input spec format", "amanzi");`

src/pks/chemistry/CMakeLists.txt
ecoon: No, it isn't redundant.  Lots of Amanzi tests are still valid ATS tests, and the fact that this is a unit test and not a regression test suggests that it ought to work.  The `if ...` makes it work, the comment explains why it is necessary and tells the developer not to remove it.

src/pks/chemistry/Chemistry_PK.cc
ecoon: Yes, that is one of the key design points of the DAG.  Every PK that uses a variable should Require the field and evaluator.  Only one may call "SetComponents()" (the object, evaluator or PK, that "owns" that variable).

src/pks/transport/MDM_Isotropic.cc
ecoon: Yeah, agreed... this is pretty terrible, but it's the best I could come up with.   I've just had too many users have unit issues.  And people don't read the documentation, they assume they know what it "should" be.

src/pks/chemistry/CMakeLists.txt
ecoon: Really it should get moved to Amanzi's regression test suite.

src/pks/chemistry/Alquimia_PK.cc
lipnikov: Amanzi relies of the high-level spec; hence. layout of the native spec and its locality is not so critical as in ATS.

src/pks/chemistry/CMakeLists.txt
lipnikov: We have not enough chemistry tests and no-one is running regression suite regularly,  so it will stay to a while.

src/pks/chemistry/Alquimia_PK.cc
lipnikov: I think the easiest solution is to add to line 455 of InputConverterU.cc this:
```
plist.sublist("PKs").sublist(chemistry).remove("initial conditions");
```
I can do it tonight. Then, only ATS will have "initial conditions" sublist in chemistry.

lipnikov:
Resolve conflicts and I'll approve

lipnikov (approved):

src/pks/chemistry/Alquimia_PK.cc
ecoon: ok, that sounds good.  I'll push changes that Amanzi has no "initial conditions" list in PKs, and ATS does.

src/pks/chemistry/CMakeLists.txt
ecoon: Fair enough.

src/pks/chemistry/Alquimia_PK.cc
lipnikov: Actually, I find your solution is more robust than the condition based on a presence of "initial conditions" sublist. Anyway, "initial conditions" should go away from Amanzi's chemistry list since it is not used.

ecoon: Ok, I added this line and updated input converter tests, so this should be good now.  Please take a look when you have a minute and see if there is anything else that needs to be checked.

PR #491, opened 2020-09-06
Test for hierarchical mesh definition

lipnikov:
WhetStone needs a light-weighted mesh to implement complex hierarchical methods without heavy code templating. This is addressed here by introducing  a mini mesh class. I went a few steps ahead and split Mesh class into a serious of imbedded classes. This strategy may help to bring in and test new (internal and external) mesh frameworks as needed. I'd like to hear your opinion before moving forward with this WIP PR. The plan below describes my strategy that is tested using WhetStone package. You may ignore all changes to the WhetStone in this PR. I also removed a few obsolete member functions, as well as member functions used in very special cases. The latter were converted to stand-alone functions. The embedded hierarchy:

- StaticMeshMini. It will be used internally by WhetStone to create a surface mesh for a cell.

- StaticMeshLight. (in plans) It will be used by WhetStone. It includes additional upward and downward adjacencies.

- StaticMesh. (implemented) It adds additional necessary functionality sufficient to be used by most Amanzi PKs.
This includes but not limited to mesh maps.

- DynamicMesh (in plans) It adds mesh modifications.

- Mesh. It adds all other stuff (mesh columns).

A few additional mesh factories will support this hierarchy of meshes. The strategy will also increase portability of separate packages. For instance WhetStone can use a light mesh interface described in StaticMeshLight.


src/mesh/StaticMeshMini.hh
```diff
@@ -0,0 +1,197 @@
...
+  // On a distributed mesh, this will return all the faces of the
+  // cell, OWNED or GHOST. If ordered = true, the faces will be
+  // returned in a standard order according to Exodus II convention
+  // for standard cells; in all other situations (ordered = false or
+  // non-standard cells), the list of faces will be in arbitrary order
+  //
+  // EXTENSIONS: MSTK FRAMEWORK: by the way the parallel partitioning,
+  // send-receive protocols and mesh query operators are designed, a side 
+  // effect of this is that master and ghost entities will have the same
+  // hierarchical topology.
```
ecoon, line 50: What is a hierarchical topology?

ecoon (commented):
I'm going to have to dive into this a bit more to better understand.  You're saying that this is the hierarchy of interfaces, correct?  So where does Mesh_MSTK come in?  For instance, is there an MSTK based StaticMesh implementation?  Or is it that Mesh_MSTK is a Mesh is a DynamicMesh is a StaticMesh?  When you say "StaticMesh is implemented" is that an "as an MSTK thing" or is it "as a separate, non-MSTK thing?"

It would be good to think about how these fit with MeshLogical, MeshSurfaceCell, and MeshColumn, which could likely be replaced by one of these simpler meshes.

src/mesh/StaticMeshMini.hh
lipnikov: If I remember correctly, it states that lower-dimensional objects (edges) of a higher-dimension object (cell) are ordered the same way for master and ghost higher-dimensional object.

lipnikov:
Currently only interfaces are implemented. MSTK mesh is derived from Mesh (I would rename it to SuperMesh but this is too intrusive). It make sense to implement DynamicMesh_MSTK since Amanzi is not using columns, I think.
MeshLogical could be a StaticMesh...

ecoon:
I guess I'd like to hear a bit more about why.  Is there a disadvantage to allowing all meshes to have columns?  To allowing all meshes to be deformed?  I get having a "mini mesh" to avoid dealing with sets and maps and avoid using MSTK.  But I'm not sure why the rest are necessary.

Second, I'm very concerned with making all of the cached methods virtual.  This is why the cache was generated in the first place -- for instance, cell_volume() is in the current master, a single array access, whereas here it would be a virtual lookup.  These would definitely add up and result in a nontrivial performance hit.

lipnikov:
Why? Currently I need only MeshLight (or MeshMini) with just upward and downward dependencies and full Mesh. Looks like, this is the optimal solution.

Entity_ID_List cell_get_faces(const Entity_ID c) const;  I support but will leave this for another PR to reduce conflicts.

lipnikov:
@ecoon Please have a look at how I handled the cache member functions. If you okay, I'll clean the conflicts and change interface to cell_get_faces() as you described. This should safe us from copying from cache.

ecoon:
That's a good point on saving the copy too.  Should be able to return by const reference and just use that most of the time.  I have an Exasheds kickoff tomorrow so probably won't get to it til Friday.

ecoon:
This looks good.  

I'm not sure why things like cell_map() are virtual, but they are currently so changing it would probably be outside of the scope of this PR.  Would be nice to make, for instance, cell_map() not virtual (as all meshes currently create the map in an init stage, then just return the map on that call -- it doesn't need to be virtual).  This would let GID() be non-virtual too -- right now that should really never be called as it is pretty poor performance wise (relative to getting the map and calling the Map's GID method).  Looking more at mesh I guess I'm collecting things for a new Mesh refactor/cleanup...

src/mesh/MeshLight.hh
```diff
@@ -0,0 +1,392 @@
...
+  void cell_get_faces(
+       const Entity_ID c,
+       Entity_ID_List *faces,
+       const bool ordered = false) const {
+    cell_get_faces_and_dirs(c, faces, NULL, ordered);
+  }
+
+  // new API: cache should be build around mesh constructor,
+  // so no additional checks is needed
+  const Entity_ID_List cell_get_faces(const Entity_ID c) const { return cell_face_ids_[c]; }
```
JulienLoiseau, line 63: @lipnikov Should be returned by reference

lipnikov: Done

lipnikov:
Verbal approval to merge has been secured.

PR #497, opened 2020-10-08
minor changes for convenience and ATS coupled transport

ecoon:
This does a few minor changes:

1. Adds a new independent variable (constant) which is literally an evaluator that sets the vector to a constant value for all time and space.  This is intended to make it as simple as possible to use evaluators.

2. Adds some convenience helpers for Debugger that allows writing a single cell-based Epetra_MultiVector (the current interface only worked on CompositeVectors).  This makes it more useful for PKs that keep their own vectors.

3. Changes how the PK_DomainFunctionCoupling works.  It now expects a ncomponents + 2 vector (conserved quantity), where the first ncomponent vectors are the transported quantity and the last two are the water leaving via the domain coupling flux (e.g. surface-subsurface exchange flux) and the water at the initial time.  This allows sources and BCs to track the water in addition to the conserved quantity, making it more straightforward to deal with cases where the outgoing C may be independent of the C in the domain.  The most obvious case of this is raining e.g. NOX on a dry surface domain -- the transported NOX must have the same C as the incoming rain's C, despite the fact that the surface domain's C is 0 (because there is no water on the surface).  Less obvious cases include where, for instance, water flows from an uphill cell, onto a downhill cell, then immediately infiltrates leaving no water (and therefore 0 C) on the surface cell.  This mass would have been lost in the previous implementation.


dasvyat:
Debugging output should be removed

dasvyat:
Ethan, what is the purpose of the constant independent variable?

dasvyat:
Item 3 is the most difficult and a bit confusing. As far as I understood, you are suggesting to handle this problem in PK_DomainFunctionCoupling by providing additional information about water at the beginning of the time step.
Is it enough to simulate the propagation into subsurface of the components? What about information about sources and BCs (both water and components).
 I used a slightly different approach. TransportPK is responsible for creating and filling special copy of the transport components which is responsible for mixing.

ecoon:
> Debugging output should be removed

Sure, must have missed some?  Do you remember where?  I believe that if you click on "start review" or something like that you can quote individual line(s) to point out stuff like this.

> Ethan, what is the purpose of the constant independent variable?

Effectively it is to enable lazy users (like me).  It is equivalent to a standard "independent variable" with a "function-constant" on all needed entities.  But instead of needing 15 lines of xml you need two -- the type and the value.  It just does a PutScalar() on the vector and never changes (in time).

An example usage is here: https://github.com/amanzi/ats-regression-tests/blob/ecoon/pk_cleanup/06_transport/transport-column_drainage.xml#L439

> Item 3 is the most difficult and a bit confusing

I don't think my approach is that different from yours.  I just put it into the main code instead of as a special function call, and leveraged the conserved quantity in DomainFunctionCoupling directly instead of pre-computing the "effective C".  See the attached description for detail.

[transport_coupling.pdf](https://github.com/amanzi/amanzi/files/5366444/transport_coupling.pdf)

ecoon:
I can't find any debug messages in this pull request, at least not in Amanzi code.  Let me know if you find any that I missed.

dasvyat:
Ethan, 

thank you for this detailed explanation. Sure your approach is more general and covers more cases.
I approve.

dasvyat (approved):

ecoon:
will merge this once CI passes

lipnikov:
Two days ago Daniil planned to verify a few benchmark tests. He had some concerns. Please ping him before merging.

ecoon:
@dasvyat I don't believe this should affect any existing runs -- I changed the names to keep the same "field" submodel as it was previously (for use in cases where there will always be water, e.g. fractured media), and added the new approach as a "conserved quantity" submodel.  Let me know if you have issues, but I need to merge this ASAP as it has been sitting way to long with people waiting for it to work.

PR #508, opened 2020-11-16
virtual inheritance confusion

ecoon:
There seems to be some confusion about how and when to use virtual inheritance, and that results in some significant code overlapping and re-initialization of variables.

Conceptually, we only need virtual inheritance for when we are dealing with a diamond inheritance -- eg. one PK inherits from two bases classes which each themselves inherit from the same common base class (e.g.)

    PK_PhysicalBDF --> PK_Physical --> PK
    PK_PhysicalBDF --> PK_BDF --> PK

In that case, only the latter needs to be virtual, e.g.:

    class PK_PhysicalBDF : public PK_Physical, public PK_BDF;
    class PK_BDF : virtual public PK;
    class PK_Physical : virtual public PK;

But, when this happens, _everything_ that inherits from this diamond must _manually_ construct the virtual class.

    PK_PhysicalBDF()
      : PK(),
        PK_Physical(),
        PK_BDF() {}

And this is true of things that derive from PK_PhysicalBDF too, e.g.:

    Flow_PK()
      : PK(),
        PK_PhysicalBDF() {}

There were some issues in some base classes that did not correctly use virtual (things that used virtual too often) and didn't manually call the constructor.  In general, we follow a rule of thumb where if a class has a protected variable, it should initialize that variable.  It shouldn't be the case that a base class has a variable and forces all deriving classes to initialize that variable.  Things like name, domain_name_, mesh_, vo_, and others are initialized all over the code, in some cases duplicating base class versions (e.g. domain_name_ vs domain_)  and in some cases hiding base class versions (name_ showed up a few places).

This pull request does a pass to clean a lot of this up and and fix virtual inheritance to only PK and make sure deriving PKs construct the base class PK.


src/pks/PK.hh
```diff
@@ -146,6 +163,36 @@ class PK {
...
+  void AddDefaultPrimaryEvaluator_(const Key& key) {
+    Teuchos::ParameterList elist;
+    elist.set<std::string>("evaluator name", key);
+    auto eval = Teuchos::rcp(new PrimaryVariableFieldEvaluator(elist));
+    AMANZI_ASSERT(S_ != Teuchos::null);
+    S_->SetFieldEvaluator(key, eval);
+  }
+
+  // Helper method to initialize a field
+  void InitializeField_(const Teuchos::Ptr<State>& S, const std::string& passwd,
```
lipnikov, line 177: What are the reasons behind making this function protected not public.

lipnikov (approved):
Approval is contingent to successful build.

src/pks/PK.hh
ecoon: It was protected, I just added the underscore to make clear it was protected.  In general, it isn't clear to me why it needs to be public as it isn't a part of the PK interface used by external clients, but instead gets used as an implementation detail by the PK itself.

It wasn't in PK, it used to be in PK_BDF (I think?) but it has nothing to do with BDF time integration.  Then I moved it to PK_Physical, because fields are usually associated with domains, and so to me it made sense to be only useful for a PK_Physical (which is a leaf PK that exists on a domain).  But it gets used by several of your MPCs, so that wasn't going to work.

ecoon:
@jd-moulton Does this require a new docker container?  Do the containers actually clone, or any guesses as to why this failed in bootstrap?

jd-moulton:
This is failing before the Travis-CI gets to the Docker part.  In the initialization of the Travis-CI container, it follows the same two steps that it does on our master branch:

git clone --depth=50 --branch=ecoon/pks_cleanup2
git submodule update --init --recursive

But on your branch it can't find the correct hashes for some of the submodules it pulls.  I don't understand this at the moment, but it does seem like something is inconsistent here.

ecoon:
> @ecoon This is failing before the Travis-CI gets to the Docker part. In the initialization of the Travis-CI container, it follows the same two steps that it does on our master branch:
> 
> git clone --depth=50 --branch=ecoon/pks_cleanup2
> git submodule update --init --recursive
> 
> But on your branch it can't find the correct hashes for some of the submodules it pulls. I don't understand this at the moment, but it does seem like something is inconsistent here.

So there are two issues here:

1. Why is it trying to call "git submodule update" at all?  Travis CI only builds and tests the Amanzi physics branch currently, and so does not need ATS source code or tests.  This should save some (small but nontrivial) amount of time.

2.  The hashes should be in the repos, but I can confirm and respawn another run of the CI.

jd-moulton:
@ecoon Travis-CI is being proactive in attempting to make sure it has the complete repository before it starts.   Irrelevant really for us because we are going to kickoff Docker and work inside the container.   Nevertheless, playing around a bit and searching in the repo it appears that the hash you have set for the ats-regression-tests repo 

pn1806419:ats moulton$ git submodule status
-1e28aeb2404fc9116faef5b3dc06ce27703e0863 docs/documentation/source/ats_demos
-d39b368772f130bf17d29cc0ff81ea568c43b741 testing/ats-regression-tests

does not exist in that repo.  So I think this is the only error that needs fixing.  I would update the hash and see if the push works in Travis-CI.

PR #524, opened 2021-02-24
Initialization of copies. 

dasvyat:
The major purpose of this pull request is to add the capability to initialize copies of the fields properly.
There is a bug during the restart when the main field and its copies are not equal to each other after the restart procedure.

There are also fixes for alquimia BC and source functions to be able to distinguish them from others.
Moreove, Alquimia update is not called if there is no water in a cell.

### DIFF
--- a/src/pks/chemistry/Alquimia_PK.cc
+++ b/src/pks/chemistry/Alquimia_PK.cc
@@ -63,7 +63,7 @@ Alquimia_PK::Alquimia_PK(Teuchos::ParameterList& pk_tree,
   cp_list_ = Teuchos::sublist(pk_list, pk_name, true);
 
   domain_ = cp_list_->get<std::string>("domain name", "domain");
-
+  
   // obtain key of fields
   tcc_key_ = Keys::readKey(*cp_list_,domain_, "total component concentration", "total_component_concentration"); 
 
@@ -758,7 +758,7 @@ int Alquimia_PK::AdvanceSingleCell(
 
 
   int num_iterations = 0;
-  if (alq_mat_props_.saturation > 1e-14){
+  if (alq_mat_props_.saturation > saturation_tolerance_){
          
     // Do the reaction.
     bool success = chem_engine_->Advance(dt, alq_mat_props_, alq_state_, 
--- a/src/pks/chemistry/Alquimia_PK.cc
+++ b/src/pks/chemistry/Alquimia_PK.cc
@@ -290,7 +290,6 @@ void Alquimia_PK::Initialize(const Teuchos::Ptr<State>& S)
         << S->time() << vo_->reset() << std::endl << std::endl;
   }
 
-  // S->WriteStatistics(vo_);
 }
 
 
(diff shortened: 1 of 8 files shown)


src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -585,6 +585,8 @@ void Alquimia_PK::CopyToAlquimia(int cell,
   mat_props.volume = mesh_->cell_volume(cell);
   mat_props.saturation = water_saturation[0][cell];
 
+
```
lipnikov, line 588: We typically use 1-2 empty lines to separate code blocks.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -756,16 +758,20 @@ int Alquimia_PK::AdvanceSingleCell(
...
-  int num_iterations;
-  bool success = chem_engine_->Advance(dt, alq_mat_props_, alq_state_, 
-                                       alq_aux_data_, alq_aux_output_, num_iterations);
-  if (not success) {
-    if (vo_->os_OK(Teuchos::VERB_MEDIUM)) {
-      Teuchos::OSTab tab = vo_->getOSTab();
-      *vo_->os() << "no convergence in cell: " << mesh_->cell_map(false).GID(cell) << std::endl;
+
+  int num_iterations = 0;
+  if (alq_mat_props_.saturation > 1e-14){
```
lipnikov, line 763: 1. I recommend to define all numerical constants in a header or at least the class top, if they're internal constants. 
2. I do not see a similar if-statement for cell initialization at time "0". Why?
3. Alquimia PK should be aligned to Native PK in this regard. If you restrict something here, do it also in Native PK.
4. Should the cut-off scenario be s=0? 
5. MOST IMPORTANT. If s=1e-14 considered critical for CrunchFlow, or PFloTran, and/or PHREEQC then it should be probably handled inside Alquimia.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -797,6 +803,12 @@ bool Alquimia_PK::AdvanceStep(double t_old, double t_new, bool reinit)
     prev_time_step_ = dt;
   }
 
+  if (vo_->os_OK(Teuchos::VERB_MEDIUM)) {
```
lipnikov, line 806: This should be probably VERB_EXTREME, since the message has little information.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -959,6 +972,7 @@ void Alquimia_PK::CommitStep(double t_old, double t_new, const Teuchos::RCP<Stat
 {
   if (S_ != S) CopyFieldstoNewState(S);
   saved_time_ = t_new;
+
```
lipnikov, line 975: Most functions in this file do not use an empty line at the end. I recommend to be consistent.

lipnikov (commented):
I posted one critical comment.  I'll delay my approval b/c to it.

src/pks/chemistry/Alquimia_PK.cc
dasvyat: I've added the saturation tolerance parameter and added similar changes to Amanzi_PK
At time "0" process is different since it doesn't depend on dt.

src/pks/chemistry/Alquimia_PK.cc
dasvyat: Fixed

src/pks/chemistry/Alquimia_PK.cc
lipnikov: Can you clarify why chemistry restrictions should be handled by Amanzi PK and not by Alquimia TPL?

dasvyat: Sergi would explain it maybe in more detail. The problem in Alquimia, especially on the surface, when a cell almost completely dry out, the time step is relatively large then Alquimia can't converge.

lipnikov: I do not see the new changes. Did you push?

lipnikov (approved):

PR #526, opened 2021-03-02
ajk: diagonal shift, multisubgrid

ajkhattak:
1. Previously diagonal shift, to avoid nonsingular Jacobian, was applied globally to the matrix. In this PR, we apply a small shift (1.e-6) to non-zero diagonal term and a relatively bigger shift (which is need in highly nonlinear surface-standalone flows and specified by the user). This approach works very well for our Arctic multiscale model.
2. PK_DomainFucntionSubridReturn is extended to **multi-subgrid** stream-hyporheic transport models.


src/data_structures/MatrixFE.cc
```diff
@@ -139,7 +139,10 @@ MatrixFE::DiagonalShift(double shift) {
   int ierr(0);
   Epetra_Vector diag(RowMap());
   ierr = matrix_->ExtractDiagonalCopy(diag);
-  for (int i=0; i!=diag.MyLength(); ++i) diag[i] += shift;
+  for (int i=0; i!=diag.MyLength(); ++i) {
```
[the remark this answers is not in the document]
ajkhattak: I see...  I can try your suggestion DiagonalShiftAdaptive(shift, value) too.

lipnikov (approved):

src/data_structures/MatrixFE.cc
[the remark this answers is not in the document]
ecoon: I'm a little concerned that this results in step discontinuities -- small changes in diag[I] result in shift_major - shift_minor changes in the shift.

Would something like:

```
DiagonalShiftFloor(double floor) {
  diag[I] = max(diag[I], floor);
}
```
do what you're looking for?  This seems safer to me.

src/pks/PK_DomainFunctionSubgridReturn.hh
```diff
@@ -136,7 +143,7 @@ void PK_DomainFunctionSubgridReturn<FunctionBase>::Compute(double t0, double t1)
       // find the subgrid gid to be integrated
       auto gid = map.GID(*c);
       std::stringstream domain;
-      domain << "subgrid_" << gid << "-";
+      domain << field_out_prefix_ << ":" << gid << "-";
```
ecoon, line 146: Don't compose these manually please -- use the functions in Keys `Keys::getDomainInSet(field_out_prefix_,gid)`

But this class should get the DomainSet object from State, and iterate over that rather than hard-coding the name anyway.  What happens when you want two sub grid models on a secondary river, and one sub grid model on a primary stream?  Or only use subgrid models on a single reach?  You should be able to define DomainSets on arbitrary subsets of the mesh, not just "all cells" like is hardcoded here.  So rather than looping over cells, the pattern should be:

1. get the DomainSet name from the parameter list
2. At some point, get the DomainSet from State
3. loop over domains in the set:

```
for (const auto& domain : domain_set) {
  int gid = Keys::getDomainSetIndex<int>(domain);
  int c = map.LID(gid);
  // do stuff
}

File: src/state/State.cc [hunk 3]
```diff
@@ -58,6 +58,7 @@ State::State(const State& other, StateConstructMode mode) :
     field_factories_(other.field_factories_),
     time_(other.time_),
     position_in_tp_(other.position_in_tp_),
+    domain_sets_(other.domain_sets_),
```
ecoon, line 61: Thanks for fixing this!

src/pks/PK_DomainFunctionSubgridReturn.hh
```diff
@@ -69,6 +69,13 @@ void PK_DomainFunctionSubgridReturn<FunctionBase>::Init(
   // get and check the model parameters
   Teuchos::ParameterList blist = plist.sublist("source function");
   field_out_suffix_ = blist.get<std::string>("subgrid field suffix");
+
+  // check for prefixing -- mostly used in the multisubgrid models
+  if (blist.isParameter("subgrid field prefix"))
```
ecoon, line 74: This parameter should be called "subgrid domain name" or "subgrid domain set".  The user would provide either "subgrid:*" or "subgrid".

ecoon (changes requested):
See comments -- big one is the switch to use the DomainSet rather than hard-code iterating over all cells.

src/pks/PK_DomainFunctionSubgridReturn.hh
```diff
@@ -139,31 +141,30 @@ void PK_DomainFunctionSubgridReturn<FunctionBase>::Compute(double t0, double t1)
...
-      for (int c_sg=0; c_sg!=ncells_sg; ++c_sg) {
-        for (int k=0; k!=nfun; ++k) {
-	  val[k] += vec_c[k][c_sg] * alpha[k] ;
-        }
-      }
 
-      for (int k=0; k!=nfun; ++k) {
-        val[k] /= ncells_sg;
+      for (const auto& domain : *domain_set) {
+	// find the subgrid gid to be integrated
```
lipnikov, line 146: Missing indentation?

src/data_structures/MatrixFE.cc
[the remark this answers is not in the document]
ajkhattak: I think what I am trying to do is to cover a relatively wider range of flow gradients. A fixed smaller shift (e.g., 1E-5) works when the flow gradient between the two neighboring cells is more or less on the same order of magnitude as the shift. As the flow gradient gets bigger this fixed smaller shift does not work, especially in the case when one cell is totally dry and other is wetter, and then a major shift is needed. Looking at the logic in DiagonalShiftAdaptive, the major shift is only applied to a totally dry cell which helped based on my experience. To be more specific, the shift_minor_=1E-3 and shift_major = 10 worked well for my 100-year Arctic projections. Sorry, what is "double floor" here? I have to try it in a 100-year run before I say if it will work or not.

src/pks/PK_DomainFunctionSubgridReturn.hh
ajkhattak: This is done. Please have a look.

src/pks/PK_DomainFunctionSubgridReturn.hh
```diff
@@ -103,60 +110,64 @@ template <class FunctionBase>
...
+	for (int c_sg=0; c_sg!=ncells_sg; ++c_sg) {
+	  for (int k=0; k!=nfun; ++k) {
+	    val[k] += vec_c[k][c_sg] * alpha[k] ;
+	  }
+	}
+	
+	for (int k=0; k!=nfun; ++k) {
+	  val[k] /= ncells_sg;
+	}
+	value_[*c] = val;
```
ecoon, line 169: Good, this looks much cleaner to me.

ecoon (approved):
Let's talk about the diagonal shift quickly, but otherwise this looks much better.

src/data_structures/MatrixFE.cc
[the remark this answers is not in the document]
ajkhattak: Done!

PR #532, opened 2021-03-09
fixed restart issue related to overwriting pk initial conditions

dasvyat:
Fields read from checkpoint are marked as changed to initiate a potential update of secondary fields.

### DIFF
--- a/src/state/Checkpoint.cc
+++ b/src/state/Checkpoint.cc
@@ -227,9 +227,9 @@ double Checkpoint::Read(State& S, const std::string& file_or_dirname)
           if (solution_evaluator != Teuchos::null){
             Teuchos::Ptr<State> S_ptr(&S);
             solution_evaluator->SetFieldAsChanged(S_ptr);
-          }        
-          field->second->set_initialized();
+          }                 
         }
+        field->second->set_initialized();
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/state/Checkpoint.cc
```diff
@@ -216,10 +216,21 @@ double Checkpoint::Read(State& S, const std::string& file_or_dirname)
...
-      if (domain.empty()) domain = "domain";
-
+      if (domain.empty()) domain = "domain";      
+      
       bool read_complete = field->second->ReadCheckpoint(*output_[domain]);
-      if (read_complete) field->second->set_initialized();
+      if (read_complete) {
+        if (S.HasFieldEvaluator(field->first)) {
+          Teuchos::RCP<FieldEvaluator> fm = S.GetFieldEvaluator(field->first);
+          Teuchos::RCP<PrimaryVariableFieldEvaluator> solution_evaluator =
```
lipnikov, line 225: This type of checks is not very nice. Should we add more functions to class FieldEvaluator to keep track of the evaluator type?

dasvyat: I'm not a big fan of these checks either, but let fix this problem first and then update FieldEvaluator interfaces.

lipnikov: Agree, but create a ticket before we forget.

jd-moulton:
We just updated Alquimia and MSTK on the master branch (minor patch updates).  I merged just to force a test that builds against these new versions.

PR #537, opened 2021-03-22
Support of multiple initialization flags

lipnikov:
Design strategy is to make initialized() and set_initialized() virtual functions. Due to possibility of having multiple fields, each derived class implements different logic. So far, only Field_CompositeVector changes 
the initialization logic. Two functions that initialize particular components have been added to this class. 

Additional logic has been added to State to avoid setting initialization flags for primary evaluators. This should be done inside PKs. To support this, a new function, IsPrimary() has been added. Not the best solution, and I tend to add a field enum if there is a consensus.
  
The logic has been tested with Amanzi PK.

### DIFF
--- a/src/state/FieldEvaluator.hh
+++ b/src/state/FieldEvaluator.hh
@@ -86,10 +86,10 @@ class FieldEvaluator {
           Key requester, Key wrt_key) = 0;
 
   // Update the field if needed.
-  //  virtual UpdateField(const Teuchos::Ptr<State>& S) = 0;
+  // virtual UpdateField(const Teuchos::Ptr<State>& S) = 0;
 
   // Update the field's derivative with respect to wrt_key if needed.
-  //  virtual UpdateFieldDerivative(const Teuchos::Ptr<State>& S, Key wrt_key) = 0;
+  // virtual UpdateFieldDerivative(const Teuchos::Ptr<State>& S, Key wrt_key) = 0;
 
 
   virtual bool IsDependency(const Teuchos::Ptr<State>& S, Key key) const = 0;
@@ -100,6 +100,9 @@ class FieldEvaluator {
   virtual std::string WriteToString() const = 0;
 
   friend std::ostream& operator<<(std::ostream&, const FieldEvaluator&);
+
+  // special flags
+  virtual bool IsPrimary() { return false; }
 
  protected:
   // parameter list for the object
--- a/src/state/FieldEvaluator.hh
+++ b/src/state/FieldEvaluator.hh
@@ -54,6 +54,7 @@ registered with the evaluator factory.
 #include "Teuchos_RCP.hpp"
 
 #include "State.hh"
+#include "StateDefs.hh"
 
 namespace Amanzi {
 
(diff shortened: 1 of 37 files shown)


src/state/FieldEvaluator.hh
```diff
@@ -101,6 +101,9 @@ class FieldEvaluator {
 
   friend std::ostream& operator<<(std::ostream&, const FieldEvaluator&);
 
+  // special flags
+  virtual bool IsPrimary() { return false; }
```
ecoon, line 105: I think it makes more sense to make a distinction of the variable type (e.g. via enum) between primary, secondary, and independent variables, e.g. `virtual Evaluator_type get_type()` or something similar.  But I'm ok with IsPrimary().

src/state/Field_CompositeVector.cc
```diff
@@ -623,8 +625,34 @@ long int Field_CompositeVector::GetLocalElementCount() {
...
 }
 
-// void Field_CompositeVector::CopyFace2BndFace(const Epetra_MultiVector& vec_face, Epetra_MultiVector& vec_bndface) {
-//   for (int bf=0   
-// }
+
+bool Field_CompositeVector::initialized() const {
+  bool flag(true);
+  for (auto name = data_->begin(); name != data_->end(); ++name) {
+    if (*name != "boundary_face") {
```
ecoon, line 632: Why not boundary_face?  There are times when this must be initialized too, and we have had bugs where cells were inited by boundary_face was not, which resulted in strange errors.

src/state/State.cc
```diff
@@ -1007,7 +1007,7 @@ void State::InitializeEvaluators() {
   }
   for (evaluator_iterator f_it = field_evaluator_begin(); f_it != field_evaluator_end(); ++f_it) {
     f_it->second->HasFieldChanged(Teuchos::Ptr<State>(this), "state");
```
ecoon, line 1009: This call to HasFieldChanged() should probably be inside the conditional.  No need to call on a primary field evaluator, it basically does nothing.

ecoon (approved):
This looks very useful to me.  I'll have to add a ticket to ATS to leverage this, as we have several initializers that set one component but not all (e.g. faces from cells, or cells from subsurface parent mesh faces, etc).

src/state/Field_CompositeVector.cc
ecoon: Specifically Manning's coefficient needs to be set on boundary_face, so this will be useful for detecting when the user forgot to do this (this is a note to myself)...

src/state/FieldEvaluator.hh
lipnikov: I'd like to have an enum instead of IsPrimary(). I'll add it to this PR.

src/state/Field_CompositeVector.cc
lipnikov: The problem is with the vandeley component which is created on a fly from a face component. I did not resolve all challenges.

src/state/State.cc
lipnikov: enum will help here.

PR #543, opened 2021-03-25
Refactory of native chemistry kernel

lipnikov:
Summary of changes:
(a) Switched to shared pointers for better memory tracking
(b) Removed dozen of unused functions
(c) Switched to LAPACK in chemistry solvers. Comment: since control of Jacobian properties has to be done outside of linear solver, there is no need to duplicate LU algorithms.
(d) Aligned code with the rest of Amanzi.
(e) Added initial functionality for enforcing geochemical constraints.
(f) Removed typedefs for int and std::string


jd-moulton:
@lipnikov have you built the user guide with this updated chemistry?

lipnikov:
Thermodynamic data base was converted to an XML format, but regression tests were not updated. I was using reference data from these tests to monitor correction of changes.

smolins:
I cannot build with structured code. It is looking for a beaker.hh file needed in src/structured_grid/mpc_pk/PM_setup.cpp:44:10

How do you want to deal with the structured code? You want to disable any native chemistry there given we don't use it?

lipnikov:
Looks like a recent merge issue with master. The structured component was tested before. Fixed

smolins:
While it is great to have the new xml database option, there are 2 issues associated with that I would resolve before merging:

- I would discontinue the use of the old .gdb format file. Having both options is confusing as demonstrated by the chemistry benchmark problems where .gdb file is listed in the top but then a database xml block available. So I am not sure which one is used
- While all these examples we have are based on a short database, there is typically one single database that contains 1000s of reactions. The new xml format allows for flexibility, so that if needed the xml DB can go in the input file but generally it would be good if the option of an external file with the same format was available. Such an option would replace the .gdb file

smolins:
The amanzi-u-1d-isotherms.xml benchmark problem is an example of two things to be careful with

- the concentrations are given both in the IC and BC block and in the constraint block. The constraint block is not used as the name is not used anywhere but it may be confusing. Most importantly, in one case (IC/BC) the concentrations are total concentrations while in constraint they are given as constraints which in some cases it means they can also be total concentrations. I like the flexibility but this needs to be written clearly in the user manual.
- the parameters for some reactions such as isotherms may be given in the materials block and a reaction parameter. The material block overwrites as it should because it provides fine control on the parameters. This is also an issue with Alquimia where one may want to rely on engine-provided parameters but with Alquimia one would have to be explicit. One option for both this here and Alquimia is to be able to not require the parameter values as material parameters always. Not sure if this is done here or the code will require them.

lipnikov:
Extraction of reactions from a big database should be done manually now.

Issue #546, opened 2021-03-26
User needs to specify path to blas and lapack libraries 


jd-moulton:
We use a very old FindBLAS and FindLAPACK that have been hacked for some special cases but are no longer covering our users needs.   The biggest issues arise on RedHat and CentOS systems and clusters with no system blas or lapack and user installed versions.  Also, support for linking to the MKL with recent intel compilers has improved in updated releases of FindBLAS and FindLAPACK.   

We will explore using the new releases and determine if any of our previous hacks are still needed or we can work with the stock release now.

jd-moulton:
First cut (in commit a90436bb8) adds the following options to bootstrap.sh:

    --with-blas=<PATH to BLAS LIBRARIES>
    --with-lapack=<PATH to LAPACK LIBRARIES>

Which is passed through to CMake with something like

    -DTPL_BLAS_DIR=<PATH to BLAS LIBRARIES>
    -DTPL_LAPACK_DIR=<PATH to LAPACK LIBRARIES>

The latest versions of FindBLAS.cmake and FindLAPACK.cmake were copied from CMake 3.20.0, and updated to search only the BLA_DIR and LAPCK_DIR respectively if set by the user.

jd-moulton:
Added an option to output debugging information for the BLAS/LAPACK find_package operations, and to pass the BLAS vendor name.  In bootstrap.sh these are simply 

     --debug_find_blas
     --tpl-blas-vendor=VENDOR

where VENDOR is described in the FindBLAS.cmake readme file (e.g., All, Generic, Intel, etc.).  These are passed through to CMake as

    -DTPL_DEBUG_FIND_BLAS:BOOL=(TRUE|FALSE)
    -DTPL_BLAS_VENDOR:STRING=<VENDOR NAME>

Note it's helpful to use the BLAS vendor string when you turn debugging on as there are a lot of extraneous "vendors" that the find_package tries before it gets to the most common.

jd-moulton:
This latest update has changes in the TPLs CMake that required changes consistent changes in the Amanzi and ATS CMake builds.   Need to build special TPLs in Docker if we are going to have CI running on this branch.

jd-moulton:
Merged from PR, and working well.

Issue #442, opened 2020-06-12
Amanzi_COMMON_*FLAGS in CMAKE_*_FLAGS 


ecoon:
I believe that the intent of Amanzi_COMMON_*FLAGS in SuperBuild is for TPLs that don't use Cmake.  Is that correct?

Currently at least Trilinos and likely most Cmake-based TPL builds _also_ put Amanzi_COMMON_*FLAGS variables into the CMAKE_*_FLAGS when they call configure.  This then results in doubled up or overriding flags.  For instance, if CMAKE_CXX_FLAGS has "-O3 -DNDEBUG" and CMAKE_BUILD_TYPE=Release, then we get "-O3 -DNDEBUG -O3 -DNDEBUG" in compile lines in the TPL.  That results in a warning from the compiler, but no big deal.

In Trilinos, I provide a bootstrap option for changing the Trilinos build type.  This is extremely useful when debugging matrix assembly, Kokkos GPU code, or anything else that relies on new Trilinos features, because I want to be able to turn on Debug Trilinos and Release everythign else.  But because of the COMMON flags, we get:  "-O3 -DNDEBUG -O0 -g".  Most compilers seem to ignore the second "-O" option.

Is there a good reason for including Amanzi_COMMON_*FLAGS in CMAKE_*_FLAGS for cmake-based TPLs?

I'm going to remove it from Trilinos at least.

ecoon:
Looking at it a bit more, what I would propose to do is, for cmake-based TPLs, instead of putting CMAKE_CXX_FLAGS_{BuildType} in Amanzi_COMMON_CXXFLAGS along with CMAKE_CXX_FLAGS, instead we would pass CMAKE_CXX_FLAGS but _not_ the BuildType flags.

But I'd like to make sure this doesn't break the case you were fixing by adding the Amanzi_COMMON_\*FLAGS variables in the first place.

lipnikov:
You will need to verify manually that each CMake package will be configured correctly. I experienced all sorts of strange behavior from defaults parameters, including Debug as the default option for BuildType. This may longer no be the case, since we upgraded many packages, but I plan to return to the build system when we start looking more closely at the target-based CMake build.

ecoon:
Ok, I'll just do it for Trilinos for now then, since 1, I know Trilinos handles these flags through cmake and does so consistently with how we do, and 2, if you're planning on a bigger refactor no sense in doing a lot of work.

[...]

PR #549, opened 2021-04-20
Adds option to specify paths for BLAS and LAPACK libraries

jd-moulton:
See comments and documentation in #546.


bootstrap.sh
```diff
@@ -707,6 +737,21 @@ List of INPUT parameters
...
+      --tpl-blas-vendor=*)
+                 tmp=`parse_option_with_equal "${opt}" 'tpl-blas-vendor'`
+                 tpl_blas_vendor=$tmp
+                 ;;
+
+      --with-lapack=*)
+                 tmp=`parse_option_with_equal "${opt}" 'with-lapack'`
+                 tpl_lapack_dir=$tmp
+                 ;;
+
```
lipnikov, line 754: I remind that  we discussed to update/extend examples printed with --help.

config/SuperBuild/TPLVersions.cmake
```diff
@@ -156,6 +156,7 @@
 #   0.97.12      - update Trilinos to afc4e525 (Nov 14, 2020)
 #   0.97.13      - update Alquimia to 1.0.8
 #   0.97.14      - update MSTK to 3.3.6
+#   0.97.15      - configuration updates fixing CamelCase names.
```
lipnikov, line 159: How about 0.98.0 for the new release.

config/SuperBuild/include/Build_PETSc.cmake
```diff
@@ -174,32 +174,32 @@ ExternalProject_Add(${PETSc_BUILD_TARGET}
...
-                                          --without-valgrind
-                                          --with-cxx-dialect=C++11
-                                          ${petsc_lapack_option}
-                                          ${petsc_blas_option}
-                                          ${petsc_package_flags}
+                              ${CONFIG_PETSc_SHARED}
+                              --prefix=${PETSc_install_dir}
+                              ${petsc_mpi_compilers}
+                              --without-x
+					          --with-ssl=0
```
lipnikov, line 186: Remove TABs or additional spaces

lipnikov (approved):

config/SuperBuild/TPLVersions.cmake
jd-moulton: Probably a good idea because this breaks the build system for older code. I'll change it before the merge.

config/SuperBuild/include/Build_PETSc.cmake
jd-moulton: Good catch, we shouldn't have any tabs in our code.  We should revisit editor configuration standards.

bootstrap.sh
```diff
@@ -707,6 +751,21 @@ List of INPUT parameters
                  XSDK=TRUE
                  ;;
 
+      --with-blas=*)
+                 tmp=`parse_option_with_equal "${opt}" 'with-blas'`
+                 tpl_blas_dir=$tmp
+                 ;;
+
+      --tpl-blas-vendor=*)
```
ecoon, line 759: I'm confused by the naming, "--tpl-blas-vendor".  Wouldn't this just be "--blas-vendor" or "--with-blas-vendor"?

config/SuperBuild/CMakeLists.txt
```diff
@@ -64,6 +64,19 @@ if (BUILD_SHARED_LIBS)
...
+# Compiler Specific Flags
+# ############################################################################ #
+
+# Check for GNU compilers and add special option for gfortran version 10 and above
+
+if(CMAKE_Fortran_COMPILER_ID MATCHES GNU)
+  if (CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL "10")
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fallow-argument-mismatch")
+    set(CMAKE_Fortran_FLAGS_RELEASE "${CMAKE_Fortran_FLAGS_RELEASE} -fallow-argument-mismatch")
+    set(CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} -fallow-argument-mismatch")
```
ecoon, line 77: If it needs to be here, also please put in "CMAKE_Fortran_FLAGS_RELWITHDEBINFO".  I use this for profiling.  Does it need to be in both these and in CMAKE_Fortran_FLAGS?

ecoon (approved):
Will build and confirm it works, but looks good from a visual inspection (seeing above comments).

bootstrap.sh
jd-moulton: We have a few options specific to TPLs that start with --tpl_*, so I thought that this was reasonable.  In terms of --with-blas* this seemed to be more natural and common for file/directory paths.  I'm not sure it would make sense here.  

If we don't want to make the connection to other TPL options, then I think --blas-vendor would be good.   Does anyone have a strong opinion either way?

config/SuperBuild/CMakeLists.txt
jd-moulton: I think you're right, it doesn't need to be there.   I'll drop back to just adding to CMAKE_Fortran_FLAGS.

bootstrap.sh
ecoon: The only things I see that are --tpl options are the "whole thing" options, e.g. "--tpl-build-dir" and --tpl-config-file".  Nothing that is specific to an individual TPL.  

I would suggest --blas-vendor is better.

ecoon: I guess then the variables should also be "blas_dir" and "lapack_dir" and "blas_vendor" without the "tpl_" prefix for consistency.  We don't use variables "tpl_hypre".

config/SuperBuild/CMakeLists.txt
ecoon: I also need (for clang, I believe version 9+, but it seems almost impossible to get clang version from Apple, not Xcode versions):

--with-c-flags="-Wno-error-implicit-function-declaration"'

or HDF5 fails to compile.  I don't see it here, but I'm building TPLs now so will say for sure in a minute...

ecoon: Yeah, this fails on HDF5 for OS X Catalina or newer without 

    --bootstrap-options='--with-c-flags="-Wno-error-implicit-function-declaration"'

I can work on this later today if you don't beat me to it.

bootstrap.sh
jd-moulton: Ok, I'll make those changes later today.

config/SuperBuild/CMakeLists.txt
jd-moulton: I don't have Clang setup on my Mac, so it would be great if you could cover this case.

bootstrap.sh
ecoon: I made the changes to bootstrap, but am not sure if you have other scripts or code that depends on that interface to bootstrap that need to be updated.  If you can confirm no other places need to change, I think this is ready to merge in.

jd-moulton: Great, I'll check it now and update anything that should/could propagate to CMakeLists.txt and the build files as needed.

jd-moulton: Unless you're ok with the TPL_* prefix in our CMake stuff.

ecoon:
@jd-moulton  Can this be merged today?

jd-moulton:
@ecoon I just realized there's a problem with the structured build in Amanzi, which is perhaps a bit more fundamental to how we manage CMake_<COMPILER>_FLAGS.   But since it messes with everyone TPLs, I'm fine with you merging it tonight if you prefer.  I'll figure this out by tomorrow and either patch master directly or this PR if you haven't merged.

jd-moulton:
@ecoon I understand what is going on and the solution really depends on what control we want to have over the flags a user builds the code with, and specifically if those have any relation to the flags that we cached when we built the TPLs.

The problem is that in bootstrap.sh we pass the -DCMAKE_<LANG>_FLAGS to the CMake configuration of Amanzi and ATS even if these are empty strings.  These values that the user provided (or these empty strings) overwrite the flags coming from our TPL cache file.  So we aren't ensuring any consistency between TPL builds and the Amanzi-ATS builds.  The case that broke when we removed the apparently redundant flags from CMAKE_<LANG>_FLAGS_<CMAKE_BUILD_TYPE> was building some of the fortran routines in structured that had an argument mismatch.   Previously it got the right flag because it was preserved in the BUILD_TYPE specific flags. 

So not sure the best fix.  It relates to the Amanzi_COMMON_*FLAGS (#442).   We could cache these and use them to check on or restore things for the user.   Or we could have the user set some other flag that we append to our cached values.  Either way it seems like we don't want the user to completely override the compiler options.

jd-moulton:
@ecoon Have solution for the immediate problem.  Preserve TPL build info in TPL_CMAKE_\<LANG\>\_FLAGS, and append these to the CMAKE_\<LANG\>\_FLAGS provided by the user.   CMake string and list tools are used to remove duplicates.  Will  commit and merge today after a bit more testing.

Looks like we could right a couple of simple functions to manage duplicates within the TPL builds and also check for them here if this is still a problem with some compilers.

PR #552, opened 2021-04-28
support of lookup tables

lipnikov:
This adds support of lookup tables. Examples include water density, viscosity and internal energy. Minor refectory and renaming of the eos classes has been done to add T,p dependencies.


lipnikov:
Need to merge it due to bug fixes.

PR #559, opened 2021-05-25
Updates to better deal with DomainSets for multi-domain modeling and better variable name handling

ecoon:
<s>NOTE: all ATS tests pass, but likely this is a WIP because Amanzi tests have not been checked yet.</s>

This PR, from an Amanzi perspective, mostly affects variable naming.  It improves utilities around Keys, adds testing for Keys, and makes sure that all of Amanzi and ATS use standard naming conventions.  See Keys.hh for more documentation of that naming convention.

It also updates Alquimia to use standard names, changing things like:

 * free_ion_concentration_NO3-.cell.0 --> free_ion_concentration.cell.NO3- 
 * "total_component_concentration.cell.tracer conc" --> total_component_concentration.cell.tracer

And similar.

edit: tests all pass now for both Amanzi and ATS


jd-moulton (approved):
Looks good to me, seems reasonable to give ATS option for python.

Issue #499, opened 2020-10-08
major Mesh refactor


ecoon:
Konstantin's new changes to the Mesh infrastructure are bringing up some opportunities to refactor Mesh to make it a bit easier, more performant, and cleaner.

* With the exception of edges, we use the entirety of the cache in all runs.  Furthermore, in all cases, when we deform the mesh, we do so globally.  So there is little point to maintaining an elaborate logic around "recompute" and lazy caching.  Effectively I propose we move to a model where there is a single `UpdateCache()` call that is called both at setup and in any`Deform()` calls, etc.  Then simplify all cache methods to just return the thing instead of checking if recompute, if created, etc.

* `cell_map()` access is currently virtual.  So is `map()`.  So is `GID()`. This is bad because it means they can't be used in an inner loop.  Instead, each mesh implementation should create the maps at initialization, and the base class should have non-virtual accessors.

* Many of the methods which query topology pass in an array.  This means they must be copied -- instead, one could return those arrays by const reference (since they are cached as arrays already).  For instance, `void cell_get_faces(Entity_ID c, Entitity_ID_List& faces)` could become `const Entitity_ID_List& cell_get_faces(Entity_ID c)`.  

* Finally, the container used in the Cache may need to change -- for instance, we have a version of this in Tpetra branch that uses Kokkos::View<Entity_ID*> instead of std::vector<Entity_ID>.  Some flexibility here would help a lot.

lipnikov:
I support simplifying the lazy caching logic. Also, with UpdateCache() we should be able to eliminate virtual *internal_() functions from MeshLight making is more "light". 

In dynamic cases, having two meshes may be useful. I suggest that major refectory will make a room for a mesh copy constructor.

ecoon:
Ok, I've started in on this a bit.  First task is splitting this beast up a bit.  I don't believe that the following capability should be in Mesh:
- deform should be nonmember functions that act on a mesh, but shouldn't be a part of the main interface
- sets (with the exception of LabeledSets) should not be in the derived mesh class, but instead outside of it, which would allow that functionality to be used on all meshes
- columns should not be a part of Mesh, but outside of it
- conflating the cache and Mesh itself has caused a huge bloat of code
- The "recompute" option is not used in Amanzi, and the "ordered" option is trouble since it falls back to something expensive. 

I'm proposing the following:

[...]

PR #567, opened 2021-06-27
updates CMake and bootstrap to standardize naming of mesh frameworks 

ecoon:
All CMakeLists options now use:  ```MESH_<FRAMEWORK> ```

and correspondingly ```ENABLE_MESH_<FRAMEWORK>.```

Previously, all source code and tests were named this way, e.g. Mesh_MSTK.hh, and the classnames were the same, but the CMake project/component was just plain MSTK (which wasn't good because we had both, e.g. MSTK_INCLUDE_DIR which referred to the TPL and MSTK_SOURCE_DIR which referred to the folder src/mesh/mesh_mstk.  Also directories are named this way, e.g. src/mesh/mesh_mstk.

But libraries were called "mstk_mesh.so" and "ENABLE_MSTK_Mesh".  Standardizing is a good thing.

This is a first step on the way to a refactor of AmanziMesh testing, and then to refactoring AmanziMesh itself.  Refs #499


ecoon:
Note @jd-moulton  this breaks TPL builds because the CMakeCache variables ENABLE_MSTK_Mesh are changed.

lipnikov:
STK should go away, since we removed it from the code.

bootstrap.sh
```diff
@@ -1924,12 +1908,10 @@ cmd_configure="${cmake_binary} \
     -DENABLE_KOKKOS_OPENMP:BOOL=${kokkos_openmp} \
     -DENABLE_AmanziPhysicsModule:BOOL=${amanzi_physics} \
     -DENABLE_ATSPhysicsModule:BOOL=${ats_physics} \
-    -DENABLE_DBC:BOOL=${dbc} \
     -DBUILD_SHARED_LIBS:BOOL=${shared} \
     -DCCSE_BL_SPACEDIM:INT=${spacedim} \
     -DENABLE_Regression_Tests:BOOL=${reg_tests} \
-    -DMPI_EXEC_GLOBAL_ARGS:STRING=${mpi_exec_args} \
-    -DPYTHON_EXECUTABLE:STRING=${python_exec} \
+    -DMPI_EXEC_GLOBAL_ARGS:STRING=${mpi_exec_args}\
```
lipnikov, line 1914: space is missing "\"

src/mesh/CMakeLists.txt
```diff
@@ -54,14 +54,23 @@ add_install_include_file(${mesh_inc_files})
 add_subdirectory(mesh_simple)
 
 # MOAB Mesh framework
-if (ENABLE_MOAB_Mesh)
+if (ENABLE_MESH_MOAB)
+  message(ERROR "==================================================")
```
lipnikov, line 58: Why "ERROR". Enabling framework is not an error.

CMakeLists.txt
```diff
@@ -88,14 +88,14 @@ endif()
 
 # Check the mesh framework choice
 if ((ENABLE_Unstructured)  AND
-    (NOT ENABLE_STK_Mesh)  AND
```
lipnikov, line 91: STK was deprecated long time ago. We should stop upgrading relevant pieces on the code.

lipnikov (commented):
Looks innocent. Since CI is failing, someone has to confirm that all tests have passed.

bootstrap.sh
ecoon: Looks like this didn't merge?  This reverts a change I made recently -- will fix.

ecoon: Yeah, ok, so I messed up a merge.  This is fixed in the next commit.

src/mesh/CMakeLists.txt
ecoon: > STK should go away, since we removed it from the code.

I've removed all references to STK that I can find now.

CMakeLists.txt
ecoon: I've now removed all references to STK that I can find, including in Trilinos configuration, etc.

ecoon:
Note, this does not pass MOAB tests for me, but neither does master currently.  I'm  not sure why -- it is cannot load the test meshes correctly, and is looking for a "GLOBAL_ID" attribute that is not in the file.  There is an "ID" attribute in the file, but no "GLOBAL_ID."   Do these tests pass for others? (Note that building without MOAB is the default, so I haven't been running those tests lately by mistake.)

lipnikov:
FYI: MOAB tests pass for me on master.

ecoon:
> FYI: MOAB tests pass for me on master.

If I run the current master, in CMAKE_BUILD_TYPE=Release, I get:
```
❯ ./moab_mesh_serial
Problem getting tag handle for GLOBAL_ID
Problem getting tag handle for GLOBAL_ID
Problem getting tag handle for GLOBAL_ID
Success: 3 tests passed.
Test time: 0.04 seconds.
```

If in Debug, I get:

```
❯ ./moab_mesh_serial
Problem getting tag handle for GLOBAL_ID
Assertion failed: (false), function ErrorCheck_, file /Users/Shared/ornldev/code/ats/amanzi/repos/develop/src/mesh/mesh_moab/Mesh_MOAB.hh, line 248.
```

https://github.com/amanzi/amanzi/blob/master/src/mesh/mesh_moab/Mesh_MOAB.hh#L248

Note that line 248 of Mesh_MOAB.hh is a C assert call, NOT an AMANZI_ASSERT, so it gets hidden for -DNDEBUG (which is disabled on CMAKE_BUILD_TYPE=Release by default, unlike AMANZI_ASSERT).  

The failure is in Mesh_MOAB.cc:init_global_ids():
```
   316 	    result = mbcore_->tag_get_handle("GLOBAL_ID", 1,
   317 	                                    MB_TYPE_INTEGER, gid_tag,
   318 	                                    MB_TAG_CREAT|MB_TAG_DENSE, &tagval);
-> 319 	    ErrorCheck_(result, "Problem getting tag handle for GLOBAL_ID");
```
where Mesh_MOAB is trying to create a new tag, but when I run it through the debugger this tag already exists, and so the result value is "MB_ALREADY_ALLOCATED."


@lipnikov can you confirm that this test passes for you in Debug build without -DNDEBUG and with no printed "Problem getting tag handle for GLOBAL_ID" statement?

lipnikov:
I see the same error messages. So it is consistent. Since MOAB is used in Amanzi tests, this error does not lead to code crash. I suggest to make a ticket if you do not want to dig into MOAB and/or Amanzi logic.

ecoon:
@jd-moulton All tests pass for me now, including moab tests.  This breaks CI because of changing ENABLE_MSTK_Mesh --> ENABLE_MESH_MSTK, etc, so TPLs for CI will have to be rebuilt.  Do you want me to merge this PR or leave it to you to merge when you're ready to update CI?

ecoon:
> I see the same error messages. So it is consistent. Since MOAB is used in Amanzi tests, this error does not lead to code crash. I suggest to make a ticket if you do not want to dig into MOAB and/or Amanzi logic.

#570

ecoon:
Something got really messed up in git on these commits.  Some things got merged, then unmerged, then lost.  Currently we have a really messed up state on master, and I don't understand why.

For instance, the option ENABLE_MSTK_Mesh was nominally changed to ENABLE_MESH_MSTK.   I went through at one point with find and made sure there was no more instances of ENABLE_MSTK_Mesh anywhere in the code.  In current master, we have, e.g. in bootstrap:

https://github.com/amanzi/amanzi/blob/master/bootstrap.sh#L1785

the correct change.

But in AmanziOptions, the change is not there:

https://github.com/amanzi/amanzi/blob/master/config/SuperBuild/CMakeLists.txt#L197

I don't see how this could possibly build at the moment...

jd-moulton:
Not sure how I messed this up.   I did make an update to a patch on Alquimia's FindPETSc.cmake but not sure how that merge messed things up.  So no overlap here. I can look into this today, we should probably rebuild from your branch and remerge any differences.

ecoon:
I don't think you messed it up -- I don't see any commit touching that line via blaim, so it must have been somehow I didn't push something or had something in a different branch or some weird mistake in git management.  I know for sure that I fixed that line, but I also managed to lose stuff in a merge.  Probably I messed it up then.

jd-moulton:
Ok, just let me know when you think this is ready for testing again and I can update the TPLs on CI.  I'll keep the version the same, and just replace the image that's there now.

ecoon:
Rebuilding TPLS from scratch, then will push once all goes through (and there is no ENABLE_MSTK_Mesh in my CMakeCache...)

ecoon:
Note I will do it as a new PR in a new branch.

PR #571, opened 2021-06-29
Fixed second link in INSTALLmd file

amaroig:
Previous link was broken


ecoon:
Thanks for pointing this out.  Also need to change the link -- will merge this in and do that change now.

PR #574, opened 2021-07-01
fixes issues remaining in failed merge/bad commits on mesh_refactor_prequel

ecoon:
See discussion in #567


ecoon:
@jd-moulton  one more fix incoming -- this is still not right...

ecoon:
@jd-moulton Ok, this builds from scratch on CADES and on my laptop.  Note you don't _have_ to build from scratch -- rerunning bootstrap is sufficient to update the cache correctly.  There are extra variables in that case, but it doesn't force users to complete rebuild TPLs.

PR #575, opened 2021-07-01
Updates to apply assembled operators when possible

ecoon:
NOTE: this PR is blocked by #574 but has already merged in that branch.  So hopefully when that PR is closed, this will merge in nicely.  To see what is actually in this PR, just look at the first commit.  All others are merges.

This PR is conceptually quite simple -- it aims to use assembled matrices in *Operator::Apply() when those assembled matrices are already available.  This can be triggered in two ways -- either for cases where the matrix was assembled in the course of doing an ApplyInverse(), e.g. the preconditioner is an assembled PC, or in the case where the user/developer has manually called AssembleMatrix().

When it was chosen to have the default forward Apply of an operator use the "unassembled" apply, it was checked and confirmed that "assemble + apply once" was more expensive than "unassembled apply once."   This is not true, however, if the Apply() operation is done more than once (I have not tested, but suspect the apply could only be done twice and already the assembled form would be cheaper).  In particular, Krylov space methods may result in Apply being called many times before changing values in the matrix.  Furthermore, in the case of Krylov space methods, if the PC is the same as the operator, the forward matrix has already been assembled, meaning that even one ApplyAssembled() is likely cheaper than one ApplyUnassembled().

The implementation is a bit tricky, however, because we have to keep track of whether the matrix has been changed since it was last assembled, and because options like "diagonal shift" to keep a PC invertible should not be used in the forward matrix.

All tests pass.  Timing checks will be done on ATS runs, but I suspect this will be a significant speedup.  Comparisons of operator_diffusion_low_order test suggest that for small, serial runs with lots of Krylov space linear iterations, the speedup is over an order of magnitude.  I doubt that will be the case in large parallel runs with heavier-weight preconditioners, but will measure and find out...


lipnikov:
User guide has large-scale tests. I would question conclusions drawn from (a) small matrices like in diffusion_low_order and (b) linear problems.  Simple arguments for 3D hex mesh and lambda-system shows 36 N^3 flows (unassembled apply) and 33 N^3 (assembled apply).  Direct improvement is 8%, overhead will reduce it. I see no reasons to get a factor or 10.

ecoon:
> User guide has large-scale tests. I would question conclusions drawn from (a) small matrices like in diffusion_low_order and (b) linear problems. Simple arguments for 3D hex mesh and lambda-system in MFD discretization give 36 N^3 flows (unassembled apply) and 33 N^3 (assembled apply). Direct improvement is 8%, overhead will reduce it. I see no reasons to get a factor or 10.

First, your estimates are wrong because you're assuming you have one local operator, when nearly always we are talking about more than one (e.g. a diffusion local operator and an approximate Jacobian local operator) and often many more than one (e.g. in a multiphysics problem with other derivatives).

Second, flop estimates are meaningless when the time is dominated by face_get_cells().  I have profiled this in real applications and the mesh operations are taking at least half the time of a Mat-Vec, with the local Mat-Vec taking approximately the other half.

I don't expect to get a factor of 10 in a real problem at scale.  I didn't say I expected to get a factor of 10.  I said the only measurements I had showed a factor of 10 and that I would test bigger problems.  Please feel free to test yourself.

Currently ExaSheds is planning on launching ~500 runs on 64 cores each that take 24 hours a piece.  I would take a 5% reduction in runtime and call it a win.  I suspect it will be larger than 5%.  I have no evidence that it will be, but will measure and find out and be sure to let you know the result.

lipnikov:
I thought that face_get_cells data will be cached in the new mesh framework. I think a large Richards' test would be a useful message of  (dis)improvement of the assembled apply.

ecoon:
Yes, but the mesh refactor is not done yet and the runs will start soon.  I'll be testing surface flow + Richards, with the Newton correction turned on, along with a few other surface ODEs for transpiration and snow that will add "overhead" but probably not change things much at scale.  If you have a Richards only test at scale that would be a nice comparison too.

We can revisit after mesh refactor too.

lipnikov:
Okay, I'll try to run 2D and practical 3D Richards' tests this long WE.

ecoon:
Ok, for a integrated hydrology run including snow processes (a simple ODE) forced by real daily met data on a heterogeneous domain (20+ soil types), on 128 cores with 260K (subsurface) cells, on a pretty normal CPU architecture, considering 1022 timesteps, gnu compilers, optimized build, I have:

ApplyUnassembled (code prior to this PR): 16878 s (4.68 hours)
ApplyAssembled (code prior to this PR):  11887 s (3.30 hours)

For a speedup of nearly 30% on a production run.

I'll take the speedup, but I'm somewhat embarrassed by how slow the code is right now.  As Konstantin points out, this should theoretically not be even close to this high, but overhead introduced by either memory access patterns and bad locality and/or (more likely) some mesh implementation decisions are really slowing us down right now.

ecoon:
@lipnikov Please let me know if you still have reservations about merging this PR.

src/operators/Operator.cc
```diff
@@ -565,6 +591,8 @@ void Operator::UpdateRHS(const CompositeVector& source, bool volume_included) {
 void Operator::Rescale(double scaling)
 {
   for (auto& it : *this) it->Rescale(scaling);
+  assembly_complete_ = false;
```
lipnikov, line 594: Update of RHS should not change matrix state.

src/operators/TreeOperator.cc
```diff
@@ -167,7 +168,26 @@ int TreeOperator::Apply(const TreeVector& X, TreeVector& Y, double scalar) const
   AMANZI_ASSERT(get_domain_map().SubsetOf(X.Map()));
   AMANZI_ASSERT(get_range_map().SubsetOf(Y.Map()));
 #endif
+  if (assembly_complete_) {
+    if (shift_ != 0.0) {
+      // must "un-shift", but even still this is probably faster!
```
lipnikov, line 173: Better to write an explicit formula or detailed comment here, since implications of mangling with shift_ are hidden inside other functions.

lipnikov (approved):
Couple of minor comments... no additional review is needed.

lipnikov:
30% reduction in cost is impressive; however, the reasons are not quite clear. For instance, for the p-lambda system, cell_get_faces() is cached and the only potential problem in ApplyMatrixFreeOp() is an indirect access to memory. But indirect memory access is true for the assembled system. Jacobian uses face_get_cells() which we know is expensive. I think the integrated hydrology recalculates matrices on each NKA iteration, so the only difference is indeed
in local ApplyMatrixFreeOp vs global mat-vec.

lipnikov:
Updated crude estimates for mat-vec on cubic meshes:

p-lambda:  49N^3 vs 43 N^3  (12% reduction)
p-lambda + FV advection + cell accumulation: 62 N^3  vs 49 N^3 (21% reduction)
FV diffusion + FV  advection + cell accumulation: 25 N^3 vs 7 N^3 (72% reduction)

I estimate validity of these bounds as 3-4%.

lipnikov:
Their problem gives expected minor improvement for assembled p-lambda system: 33 sec vs 35 sec (8 cores)

src/operators/Operator.cc
ecoon: This Rescale() changes the local matrices (this->begin()/end() is an iterator over the local_op_ vector).

PR #576, opened 2021-07-02
Support for optional isotherms in material section

lipnikov:
Global sorption isotherms can be specified only in the thermodynamic database section.


smolins:
I agree with this changes and ready to approve. However, I would like to ask whether this is documented anywhere. Thanks

lipnikov:
Extended documentation of Spec 2.x and Native Spec.

PR #579, opened 2021-07-14
Hypre changes for solvers

JulienLoiseau:
- Add support for Hypre for Tpetra.


config/SuperBuild/include/Build_Trilinos.cmake
```diff
@@ -216,9 +216,12 @@ if (ENABLE_KOKKOS)
     list(APPEND Trilinos_CMAKE_ARCH_ARGS "-DKokkos_ENABLE_CUDA:BOOL=OFF")
   endif()
   if (ENABLE_KOKKOS_OPENMP)
+    message(STATUS "OPenMP activated")
```
lipnikov, line 219: Typo: OpenMP

src/data_structures/test/data_structures_CSR.cc
```diff
@@ -28,12 +28,10 @@ SUITE(COMMON_CSR)
 {
   TEST_FIXTURE(TestHarness, CSR_TEST)
   {
-    const int ncells = 1000; 
-
     // Create Matrix CSR 
     const int nmatrices = 100000; 
-    const int nrows = 10; 
-    const int ncols = 10; 
+    const int nrows = 100;
```
lipnikov, line 33: Do we need this scale level for a unit test?

src/operators/test/verify_operators.cc
```diff
@@ -183,13 +191,68 @@ TEST(Verify_Mesh_and_Operators) {
...
+    plist->sublist("preconditioners").sublist("Hypre: AMG")
+        .set<std::string>("preconditioning method", "hypre: boomer amg").sublist("hypre: boomer amg parameters")
+        .set<double>("strong threshold", 0.5)
+        .set<int>("cycle applications", 1)
+        .set<int>("smoother sweeps", 2)
+        //.set<double>("tolerance", 1e-5)
+        .set<int>("verbosity", 3)
+        .set<int>("coarsening type", 8) /* 8: PMIS */
+        .set<int>("interpolation type", 3) 
+        /*3:  direct
```
lipnikov, line 205: Cite here Hypre's manual number for the record.

src/operators/test/verify_operators.cc
```diff
@@ -183,13 +191,68 @@ TEST(Verify_Mesh_and_Operators) {
...
+        /*3: Hybrid Gauss Seidel 
+          4: ''
+          6: '' 
+          7: Jacobi  
+          18: l1-jacobi
+          11: two-stage Gauss-Seidel 
+          12: ''
+           */
+
+        //HYPRE_BoomerAMGSetAggNumLevels(precon, agg_num_levels);
```
lipnikov, line 221: ??? remove ???

src/solvers/AmanziHypreFunctionParametersWrapper.hh
```diff
@@ -0,0 +1,336 @@
+#ifndef AMANZI_HYPRE_FUNCTIONPARAMETERS_HPP
+#define AMANZI_HYPRE_FUNCTIONPARAMETERS_HPP
```
lipnikov, line 2: Looks like a Trilinos file. I find their design extremely cumbersome. I would try to avoid it as much as possible.

lipnikov (changes requested):
Large part of the Trilinos code is copied into Amanzi with some adjustments. Some adjustments involve commenting-out the original if-else-branches that should be simply removed. I think the original code is  over-desiged and we do not need such complexity. 
I would use the old approach where we call Hypre functions directly in response to a parameter in a PList.

src/solvers/PreconditionerHypre.cc
```diff
@@ -20,17 +19,45 @@
...
+
 namespace Amanzi {
+
+  #include "AmanziHypreWrapperParametersMap.hh"
+
+  void AMANZI_CHK_ERRV(int code) {
+    if(code>0) {
+      std::ostringstream ofs;
+      ofs << "Amanzi::Hypre: Error with code "<<code<<std::endl;
+      throw std::runtime_error(ofs.str());
```
lipnikov, line 37: Why not AmanziException, so the code will be terminate correctly by main.cc?

lipnikov (approved):

PR #586, opened 2021-07-29
Fix hypre build

JulienLoiseau:
- Ifpack2 is needed for its interface to Hypre (even if their current interface is very slow)


lipnikov (approved):

lipnikov:
Works with --enable-kokkos only. Did not work before.

PR #592, opened 2021-08-03
refactors mesh testing infrastructure to generalize tests

ecoon:
This PR is a second step toward a significant mesh refactor.  In that refactor, the intent is to separate the Mesh interface into a MeshFramework interface which supports generic unstructured mesh tools and a MeshCache interface which supports fast lookup of mesh topological and geometric quantities on both CPUs and devices.

This will also abstract much of the capabilities which are implemented in Mesh_* classes (e.g. Mesh_MSTK, Mesh_Simple, Mesh_MOAB) and refactor those that can be written in terms of a reduced Mesh interface to be implemented outside of the derived classes.  See #499 for more details.

Unfortunately, this refactor will break all of our testing, which makes it hard to test the refactor.  Therefore this PR is a step to first abstract the testing infrastructure around mesh objects to be:

1. more generic and based on MeshAudit where possible
2. less duplicative in that many many tests do the same things over and over again on slightly different meshes, or even the same mesh
3. more abstract and using only the Mesh interface with no knowledge of the underlying framework.

The intent is that the tests will first be refactored into this new paradigm (e.g. test based on Mesh not based on Mesh_MSTK/Mesh_*) and get these tests working in the old mesh code, then refactor the mesh code with confidence thanks to this change in tests.

This passes all tests for me, but is still listed as a WIP because likely there are a few more tests to remove or refactor (e.g. pinchout tests in MSTK).

Edit: Have now removed all MOAB and Simple tests after examining each one to make sure there was nothing being tested there that was not tested via the Mesh interface in mesh_geometry and mesh_sets tests.  The MSTK tests need a bit more looking, because there are pinchout tests there (which will get added to mesh_geometry) and a test on face inheritance on surface meshes, which is tricky and needs careful inspection to see if I have covered all cases tested there in the mesh_surface_extraction test.


src/mesh/mesh_extracted/MeshColumn.hh
```diff
@@ -89,11 +87,11 @@ class MeshColumn : public Mesh {
     Entity_ID ent;
     switch (kind) {
       case FACE:
-        ent = extracted_->entity_get_parent(kind, column_faces_[entid]);
+        ent = parent_mesh_->entity_get_parent(kind, column_faces_[entid]);
         break;
 
       default:
-        ent = extracted_->entity_get_parent(kind, entid);
+        ent = parent_mesh_->entity_get_parent(kind, entid);
```
lipnikov, line 94: parent mesh may have parents for some entities?

src/mesh/mesh_extracted/MeshExtractedManifold.hh
```diff
@@ -190,6 +190,11 @@ class MeshExtractedManifold : public Mesh {
   }
 
   // Mesh Sets for ICs, BCs, Material Properties
+  virtual
+  bool valid_set_type(const AmanziGeometry::RegionType rtype, const Entity_kind kind) const override {
+    return parent_mesh_->valid_set_type(rtype, kind);
```
lipnikov, line 195: Potential problem: region may empty on the parent mesh and non-empty on the extracted mesh.

src/mesh/test/mesh_surface_extraction.cc
```diff
@@ -0,0 +1,130 @@
...
+  surf_reg_spec.set<Teuchos::Array<double>>("point", point);
+  std::vector<double> normal{0.0, 0.0, 1.0};
+  surf_reg_spec.set<Teuchos::Array<double>>("normal", normal);
+  gm->AddRegion(AmanziGeometry::createRegion("Top Face Plane", gm->size(), spec, *comm));
+
+  // a 3D, generated, structured hex on the unit cube, NX=NY=NZ=3
+  // works in MSTK
+  //
+  // extract & flatten the top surface to form a 2D mesh, NX=NY=3
+  std::vector<Framework> frameworks;
```
lipnikov, line 51: Alternative extraction works for most parent mesh frameworks. Not clear if this will be tested here.

src/mesh/mesh_extracted/MeshColumn.hh
ecoon: Hmm, this raises a good point.  So previously, the implementation was that one would pass MeshColumn a volumetric mesh and a column ID.  Then we would extract the 3D column and create a sub-mesh (this was called extracted_), and then create the MeshColumn object (which is a 1D version of the column that includes N-cells and only N+1 faces, and nodes are the same as faces).  Conceptually, the extracted mesh was an intermediate step, used to implement this Mesh, but was not the "parent" -- that is the original 3D mesh.

When I moved MeshColumn out of Mesh_MSTK, I could no longer know the type of the parent mesh.  Which means I need to use the MeshFactory to construct the extracted 3D column mesh, not directly use the Mesh_MSTK constructor as it used to be.  This led to a dependency problem -- the mesh_factory library depends on the mesh_extracted library (to construct MeshExtractedManifold objects), but the mesh_extracted library then needed the mesh_factory library (to use the factory to construct the 3D column).  So to avoid the circular dependency, I changed the constructor of MeshColumn to directly take the 3D column mesh, and added a new factory (just a function) that takes the 3D volumetric mesh and a column ID.  

So yes, the parent mesh of the 1D mesh (the 3D column) has a parent (the 3D volumetric mesh).  But that is not at all obvious, and there is no reason why a 3D column of cells couldn't be passed to this constructor with no parent.

Will change a few things included the name to make this clearer.

src/mesh/test/mesh_surface_extraction.cc
ecoon: By "alternative extraction" what do you mean?  Is this MeshExtractedManifold?  Pending your response I'll try to add it to this test.

Assuming so, can you comment a bit on that implementation?  Does it use MeshCache directly?  Or just reference everything to the parent mesh?  Are the GIDs parent GIDs (and so neither sequential nor in [0,ncells) ) or are they renumbered/reordered?   I'll look at this a bit closer so I understand what this is doing.

src/mesh/mesh_extracted/MeshExtractedManifold.hh
ecoon: Resolved via 0cf7533

src/mesh/test/mesh_surface_extraction.cc
ecoon: Also, what assumptions does this make on the parent mesh?  It looks like, at the least, this requires that the parent mesh has request_edges=True.  Is that correct, and does this make the MSTK memory issue worse?

ecoon: Ok, I added a section to the test, src/mesh/test/mesh_surface_extraction that tests MeshExtractedManifold by setting the "extraction method" to "manifold mesh".  This required also updating the factory to pass the "flatten" argument to MeshExtractedManifold, which is currently missing and is always false.

The tests don't pass, but I think they aren't passing because of limitations of MeshExtractedManifold.  Specifically, face_normal(f, false, c), which should get the outward normal of a face with respect to a specific cell, does not seem to get the sign right.  So I can add the test, but it looks like there would be some more debugging needed.

lipnikov: MeshExtractedManifold relies only the Mesh class, so it expected to work for any framework. GIDs for owned maps should be contiguous, if I remember correctly.  

request_edges is required b/c functions like cell_get_faces is delegated to face_get_edges. Although, it is possible probably to build everything from minimal information, parent->cell_get_nodes, but this would be another mesh class.

data are not cached.

src/mesh/mesh_extracted/MeshColumn.hh
ecoon: Resolved via 0cf7533

src/mesh/mesh_extracted/MeshExtractedManifold.hh
lipnikov: Regarding GM, I've changed RegionBox to enforce dimensionality check. Otherwise, flattened mesh could use an unrelated (geometrically) 3D box. This messes up with logical operations. So, potentially there is a room for a 2D GM;
however, currently I have more questions than solutions.

src/mesh/test/mesh_surface_extraction.cc
lipnikov: Did you merge with master. Some minor changes has been made to this framework that may affect face_normal()

ecoon: Ok, so looking at the comment and code in cell_get_faces_and_dirs_internal_:

    // algorithms on a non-manifold use multiple normals and special continuity
    // equations for fluxes, so that orientation does not play role.
    // This may change in the future.

it appears that cell_get_faces_and_dirs_internal does not support flattened meshes, where this direction is a valid quantity.  Normals are supplied by Mesh.cc -- if manifold and space dimensions are both 2, the normal is the cross product of the edge vector, but directions are those given by MeshExtractedManifold, which are all 1, and so cannot be correct for both cells.

lipnikov: This was changed on a different branch, sorry. I can port copy changes to master.

ecoon: Haha, ok, so next I tried to add this class to the test for extracting fractures.  This should not have the problem of the surface mesh extraction test.  But the MeshExtractedManifold only works to extract one set, and most of the other frameworks don't support logical regions, so I cannot extract two fractures on a generated mesh...

lipnikov: added two code lines and two additional comment line. Should work now.

lipnikov: FYI: Our student uses a flattened mesh for coupled shallow_water and Richards. Both frameworks work there which implies that face_normal(f, false, c) works in both frameworks.

ecoon: Ok, so the test works for a generated background mesh with Simple in serial.

I don't understand how MeshExtractedManifold works in this test:
https://github.com/amanzi/amanzi/blob/master/src/mesh/mesh_extracted/test/mesh_extracted_fracture.cc#L68

given that MOAB doesn't support edges:
https://github.com/amanzi/amanzi/blob/master/src/mesh/mesh_moab/Mesh_MOAB.hh#L111

So I have commented out tests that could be fixed, and will create a ticket on issues I'm finding with MeshExtractedManifold, and you can take a look as you have time.

ecoon: See #596

ecoon: > FYI: Our student uses a flattened mesh for coupled shallow_water and Richards. Both frameworks work there which implies that face_normal(f, false, c) works in both frameworks.

Maybe it is that the student's examples only used this version of face_normal() on boundary faces and you got lucky?  It appears to work for me with your change now -- before that change, the test summed the outward normals for each cell and made sure they added to 0 -- when they didn't they printed out the cell-outward normals.  Looking at the printout it was clear the direction was not incorporated.  Was their shallow_water and Richards example calling the Mesh factory or calling the MeshExtractedManifold constructor explicitly?  Because if they were using the factory it did not correctly use the flattened parameter.

ecoon:
Note to self -- this is still WIP because of 1 remaining test in mesh_mstk that must be checked to confirm that it is covered by the test in mesh, or added into that test.

src/mesh/test/mesh_surface_extraction.cc
lipnikov: Shallow water uses face_normal(f, false, c) on all faces. Most time we use MSTK based extraction, but I tried a few tests with explicit MeshExtractedManifold constructor and found the problem with normal orientation. MOAB looks like a placeholder at the moment, the fracture test is "failing" for MOAB but does not report this failure to CMake.

lipnikov (changes requested):
nprocs=2 does not test the case where entity is a ghost entity on 2 or more processors. It would be nice to test with nprocs=3 or 4 in at least one case.

A similar special case is the case where a set has only ghost entries on a processor. Extraction based on a projection of a 3D set on manifold is not always sufficient.

ecoon:
> nprocs=2 does not test the case where entity is a ghost entity on 2 or more processors. It would be nice to test with nprocs=3 or 4 in at least one case.
> 
> A similar special case is the case where a set has only ghost entries on a processor. Extraction based on a projection of a 3D set on manifold is not always sufficient.


Done

lipnikov (approved):

ecoon:
@lipnikov Actually as I think more about this, I realize I'm not entirely covering your cases.  Specifically, I don't test getting ALL entities particularly well, as this relies on the partitioning being known.

That said, I don't think our previous tests tested this very well.  So, two questions:

1. do you use get_set_entities(ALL) often?  I don't think I do -- pretty much I follow the pattern of writing things on OWNED cells only, then scattering if needed.  (Note, this doesn't mean it shouldn't be tested, I'm just curious if you use it much.)

2. Are you content if I test this with known coloring?  I recently added a partitioning method to MSTK that allows a user to provide a coloring file and partition based on that file -- this would make it much easier to test because I could prescribe the partition and directly control what was owned/ghosted (as opposed to partitioning, then writing the test, and hoping the partitioning doesn't change in the future).

lipnikov:
Coloring sounds like a robust approach. I support it.

get_set_entities(ALL) is critical for BCs. Otherwise, I try to avoid ALL, except for old algorithms.

PR #593, opened 2021-08-03
Fix CopyTpetraToHypre

JulienLoiseau:
- Add patch to fix the CopyTpetraToHypre function from Ifpack2. 
- This will allow to remove the ifpack2 interface from Amanzi in tpetra-solvers branch.


lipnikov (approved):
I see no obvious issues.

PR #617, opened 2021-09-10
adds some simple mesh algorithms for working with boundary entities

ecoon:
Effectively this just adds the file src/mesh/Mesh_Algorithms.{hh,cc} which implement a few simple things like getting the internal cell of a boundary face, etc.

I previously had this in ATS's PK codebase, but this is a better place since they only need a Mesh object.  Does Amanzi have this type of thing in another place?  Could we combine a few of these?


src/mesh/Mesh_Algorithms.cc
```diff
@@ -0,0 +1,109 @@
...
+// -----------------------------------------------------------------------------
+// Given a face ID, and assuming it is a boundary face, get the cell internal.
+// -----------------------------------------------------------------------------
+AmanziMesh::Entity_ID
+getFaceOnBoundaryInternalCell(const AmanziMesh::Mesh& mesh, AmanziMesh::Entity_ID f)
+{
+  AmanziMesh::Entity_ID_List cells;
+  mesh.face_get_cells(f, AmanziMesh::Parallel_type::ALL, &cells);
+  if (cells.size() != 1) {
+    Errors::Message message("getFaceOnBoundaryInternalCell called with non-internal face "+std::to_string(f));
```
lipnikov, line 59: Face id is not informative in parallel. I would print out coordinates of face centroid.

lipnikov (approved):

lipnikov:
See 
(1) whetstone/WhetStoneMeshUtils.hh
(2) BoundaryFaceGetCell(f) in Flow_PK.hh, 
(3) test/operator_coupled_diffusion.cc:int BoundaryFaceGetCell

lipnikov:
@ecoon I can remove redundant functions (primarily bf->c map)  later, if you do not want to do it now.

src/mesh/Mesh_Algorithms.cc
ecoon: Good point, will fix.

ecoon:
> See
> (2) BoundaryFaceGetCell(f) in Flow_PK.hh,
> (3) test/operator_coupled_diffusion.cc:int BoundaryFaceGetCell

Got these two.

> whetstone/WhetStoneMeshUtils.hh

Let's look at these again after the mesh refactor -- it probably makes sense to move these over into Mesh_Algorithms there too.  In general I'm trying to implement more external functions like these -- since they are implemented using the Mesh interface and don't need to know internals of the mesh class, it makes sense to keep them out of the the Mesh class.

PR #623, opened 2021-10-02
Refactory of relative permeability code

lipnikov:
Converted rel perm to an evaluator. This is will support  thermal multiphase flow.


ecoon:
Please don't steal ATS code (or at least ATS names) without pull requests.  This breaks ATS because it uses the same filenames but different namespaces.

PR #632, opened 2021-11-01
Naren/shallow water RK

nvohra0016:


src/pks/shallow_water/test/shallow_water_analytical.xml
```diff
@@ -0,0 +1,161 @@
...
+      </ParameterList>
+
+      <ParameterList name="reconstruction">
+        <Parameter name="polynomial order" type="int" value="0"/>
+        <Parameter name="limiter" type="string" value="Barth-Jespersen"/>
+        <Parameter name="limiter stencil" type="string" value="cell to all cells"/>
+        <Parameter name="limiter location" type="string" value="face"/>
+        <Parameter name="limiter cfl" type="double" value="1.0"/>
+      </ParameterList>
+      <Parameter name="temporal discretization order" type="int" value="1"/>
```
lipnikov, line 37: We need to test RK2 or RK3. If we need to turn-off (temporarily) limiters for this test, we can add more parameters to XML sublist.  If you want to keep RK1, I do not mind, just convert TEST(SHALLOW_WATER_ANALYTICAL) to a function and call it with different RK parameters.

lipnikov (changes requested):

lipnikov (approved):
I verified solutions. MPI works correctly.

PR #637, opened 2021-12-03
Cleans up a few things to use communicators

ecoon:
1. PKs now create their verbose objects with the TreeVector's comm.
This is useful for domain set PKs to allow off-rank-0 comms to still
write and debug.  In particular this is useful with openmpi's
-output-filename option which will have each rank write its own file.

Note that the output of PKs on MPI_COMM_WORLD should not change.

2. Observables now get a comm.  This allows observations on vectors
that only exist on a subset of ranks.

3. Fixes bugs, race conditions, and other synchronization problems in
domain sets.

4. Adds missing check for suffixes in dependencies instead of just
keys in secondary_variable_field_evaluator.


ecoon:
This is more work toward being more careful of communicators.  The goal is moving toward the ability to spawn off multiple subdomain solves that can be linked to a referencing mesh (e.g., solving 1D columns in z in a 3D domain, or solving sub-catchment watersheds in a bigger watershed).  Each subdomain should be able to existing on a (proper) subset of MPI_COMM_WORLD and everything should still work (without MPI hanging).

ecoon:
Yuck, this got stale.  Need to move this onto konstantin/state

PR #638, opened 2021-12-16
cleanup of names, docstrings.  Brings a few concepts back from master

ecoon:
Updates Key utilities to master

Adds GetEvaluatorList() back to State from master

Some PK documentation and cleanup.  See more in corresponding ATS PR

### DIFF
--- a/src/state/evaluator/Evaluator_OperatorApply.cc
+++ b/src/state/evaluator/Evaluator_OperatorApply.cc
@@ -17,29 +17,40 @@ on a domain.
 
 It is expected to be of the form:
 
-r = rhs_1 + ... rhs_k + rhs_A00 + rhs_A01 + ... rhs_Anm - (A00 x0 + A01 x0 + ...
+
+r = b_1 + ... b_k + rhs_A00 + rhs_A01 + ... rhs_Anm - (A00 x0 + A01 x0 + ...
 A0n x0 + A11 x1 + ... Anm xn)
 
 Where:
 
-  x0                   | is the primary variable associated with this residual
-  x1 ... xn            | are off-diagonal primary variables.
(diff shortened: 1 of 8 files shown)


src/state/evaluator/Evaluator_OperatorApply.cc
```diff
@@ -90,13 +101,13 @@ Evaluator_OperatorApply::Evaluator_OperatorApply(Teuchos::ParameterList& plist)
...
-      plist_, domain, x_key_suffix + " local operator rhss", &defaults));
+      plist_, domain, x_key_suffix + " local operator rhs", &defaults));
   }
 
   // other rhss
-  rhs_keys_ = Keys::readKeys(plist_, domain, "additional rhss", &empty);
+  rhs_keys_ = Keys::readKeys(plist_, domain, "vector", &empty);
   if (rhs_keys_.size() > 0) {
-    rhs_scalars_ = plist_.get<Teuchos::Array<double>>("rhs coefficients");
+    rhs_scalars_ = plist_.get<Teuchos::Array<double>>("vector coefficients");
```
lipnikov, line 110: "vector" or "vector coefficients" ???

lipnikov (approved):

src/state/evaluator/Evaluator_OperatorApply.cc
ecoon: So OperatorApply is a summation of both the action of operators and a set of other vectors, e.g source terms and others (the b_k).  "vector keys" in the spec is an Array<string> list of their names, and this is a list of their coefficients, e.g. alpha_k * b_k.  "rhs" wasn't quite right, so this was a second try.

Definitely need to fix this variable name to be consistent, however.  Not sure how I missed that.

PR #640, opened 2021-12-28
New state for Amanzi and ATS

lipnikov:
Due to outdated branch, the new state capabilities were ported to Amanzi manually. All Amanzi tests have been upgraded to demonstrate the new capabilities are working. In addition, Tag class was introduced for tags, which revealed a few typos.


src/state/CMakeLists.txt
```diff
@@ -53,69 +57,65 @@ endif(ENABLE_Silo)
...
-  Field.hh
-  Field_Scalar.hh
+  chemistry_data.hh 
+  data/Data.hh
+  data/Data_Helpers.hh
+  data/Data_Impl.hh
+  data/DataFactory.hh
+  data/DataFactory_Impl.hh
+  data/Record.hh
+  Evaluator.hh
```
ecoon: In tpetra branch, we moved all of these Evaluator* class files into a subdirectory (evaluators).  I think that would be a good idea to do that now.

src/pks/eos/EOSDensityEvaluator.hh
```diff
@@ -15,26 +15,27 @@
 #define AMANZI_EOS_DENSITY_EVALUATOR_HH_
```
ecoon, line 15: We need to figure out if these can be shared.  I'm not sure we can share PKs yet, but we should definitely be sharing these.  Might need to check with @dasvyat as some things (at least density) are a bit different with salinity.

src/pks/flow/WRMEvaluator.hh
```diff
@@ -15,39 +15,41 @@
 #ifndef AMANZI_FLOW_WRM_EVALUATOR_HH_
```
ecoon, line 15: I didn't realize you were using evaluators in flow now.  That's great, again, we should put these in a place that can be shared, and/or figure out a naming convention to have ATS's and Amanzi's PKs coexist next.

src/state/State.hh
```diff
@@ -79,128 +80,99 @@ Example:
...
-#include "Visualization.hh"
+// Amanzi::State
 #include "Checkpoint.hh"
 #include "ObservationData.hh"
-
-#include "Field.hh"
-#include "Field_Scalar.hh"
-#include "Field_ConstantVector.hh"
-#include "Field_CompositeVector.hh"
+#include "RecordSet.hh"
```
ecoon: This does not compile for me in a clean build.  "RecordSet.hh" lives in the subdirectory src/state/data, which is not in the include_directories of State's CMakeLists.txt (so compilation fails on State.hh).

In the tpetra branch, we have `#include "data/RecordSet.hh"` rather than trying to add `include_directory(${STATE_SOURCE_DIR}/data)` into all CMakeLists.txt.

src/state/State.hh
```diff
@@ -349,24 +484,17 @@ class State {
...
-  void RequireFieldCopy(Key fieldname,  Key tag, Key copy_owner);
-  void CopyField(Key fieldname, Key pk_name, Key timetag);
-  Teuchos::RCP<const CompositeVector> GetFieldCopyData(Key fieldname, Key tag) const;
-  Teuchos::RCP<CompositeVector> GetFieldCopyData(Key fieldname, Key tag, Key pk_name);
-
   // Time accessor and mutators.
-  double time() const { return time_; }
-  void set_time(double new_time);  // note this also evaluates state-owned functions
-  void advance_time(double dT) { last_time_ = time(); set_time(time() + dT); }
+  double time(const Tag& tag = Tags::DEFAULT) const { return Get<double>("time", tag); }
```
ecoon: Should be "get_time()" I think?

src/state/State.hh
```diff
@@ -379,96 +507,43 @@ class State {
...
+  int cycle() const { return Get<int>("cycle"); }
+  void set_cycle(int cycle) { Set("cycle", Tags::DEFAULT, "cycle", cycle); }
+  void advance_cycle(int dcycle = 1) {
+    Set("cycle", Tags::DEFAULT, "cycle", Get<int>("cycle") + dcycle);
+  }
 
   // Position accessor and mutators.
-  int position() const { return position_in_tp_; }
-  void set_position(int pos ) { position_in_tp_ = pos; }
+  int position() const { return Get<int>("position"); }
```
ecoon: and get_position()

src/state/Tag.hh
```diff
@@ -0,0 +1,56 @@
...
+
+#include "Key.hh"
+
+namespace Amanzi {
+
+class Tag {
+ public:
+  Tag() : tag_("") {};
+
+  void set(const Key& key) { tag_ = key; }
```
ecoon: Shouldn't this accept a std::string and not a Key?

src/state/Tag.hh
```diff
@@ -0,0 +1,56 @@
...
+#ifndef AMANZI_STATE_TAG_HH_
+#define AMANZI_STATE_TAG_HH_
+
+#include "Key.hh"
+
+namespace Amanzi {
+
+class Tag {
+ public:
+  Tag() : tag_("") {};
```
ecoon: Shouldn't there be a constructor from const char*?  Are you intentionally excluding the usage:

S->Require<...>("myfield", "mytag");

src/state/Tag.hh
```diff
@@ -0,0 +1,56 @@
...
+
+  void set(const Key& key) { tag_ = key; }
+  std::string get() const { return tag_; }
```
ecoon: Types of get() return value, set() argument, and tag_ should all match.  I think they should be strings and not Keys.

src/utils/Key.hh
```diff
@@ -210,6 +211,34 @@ Key readKey(Teuchos::ParameterList& list,
...
+// Convenience function to see if a map (or map-like) object has a key.
+template<typename T, typename K>
+bool hasKey(const T& container, const K& key) {
+  return container.count(key) > 0;
+}
+
+// Tag'd variables are of the form VARNAME:TAG
+inline Key
+getKeyTag(const Key& var, const Key& tag) {
+  return tag.empty() ? var : var+":"+tag;
```
ecoon: ":" can't be used here -- it is reserved for domain sets (see the documentation in this file's header).  I would suggest setting a delimiter char at the top in the same way that the rest of the delimiters are set, then use `merge(var, tag, tag_delimiter)`

src/utils/Key.hh
```diff
@@ -210,6 +211,34 @@ Key readKey(Teuchos::ParameterList& list,
...
+}
+
+// Split a DOMAIN-VARNAME key.
+inline KeyPair
+splitKeyTag(const Key& name)
```
ecoon: This should call `Keys::split(name, tag_delimiter)` instead of the current implementation.

ecoon (changes requested):
A couple changes here, some questions.  

A key question seems to be whether either of these are valid:

    S->Require<CV,CVS>("mykey", "myowner");  // the equivalent of this is valid in master
    S->Require<CV,CVS>("mykey", "mytag");      // currently this is valid in tpetra branch

I believe that currently in this PR, neither are.  Should either be valid?  Clearly both can't be.


Also, do you have a sed script that you used to do your Require<> and Get<> calls?

src/data_structures/CompositeVector.hh
```diff
@@ -171,20 +171,21 @@ class CompositeVector {
   name_iterator begin() const { return map_->begin(); }
   name_iterator end() const { return map_->end(); }
   unsigned int size() const { return map_->size(); }
+
   Comm_ptr_type Comm() const { return map_->Comm(); }
   Teuchos::RCP<const AmanziMesh::Mesh> Mesh() const { return map_->Mesh(); }
   bool HasComponent(std::string name) const { return map_->HasComponent(name); }
   int NumComponents() const { return size(); }
-  int NumVectors(std::string name) const { return map_->NumVectors(name); }
```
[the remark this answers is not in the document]
lipnikov: Done.

src/data_structures/CompositeVector.hh
```diff
@@ -382,7 +383,7 @@ class CompositeVector {
   // -- Utilities --
 
   // Write components to outstream.
-  void Print(std::ostream &os, bool data_io = true) const;
+  void Print(std::ostream& os, bool data_io = true) const;
```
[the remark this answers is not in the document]
lipnikov: There is no consistency. I follow Amanzi codying standard: "Assignment operators always have spaces around them."

src/mpc/test/mpc_ihm_shallow_water_transport.cc
```diff
@@ -57,7 +57,7 @@ using namespace Amanzi::AmanziGeometry;
   cycle_driver.Go();
   
   // verify the maximum principle for solute transport
-  const Epetra_MultiVector& tcc = *S->GetFieldData("total_component_concentration")->ViewComponent("cell");
+  const auto& tcc = *S->Get<CompositeVector>("total_component_concentration").ViewComponent("cell");
```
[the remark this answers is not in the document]
lipnikov: Yeah, I changed many of them for this reason.

src/mpc/test/observable_line_segment.cc
```diff
@@ -69,14 +69,14 @@ TEST(OBSERVABLE_LINE_SEGMENT) {
   Teuchos::RCP<State> S = Teuchos::rcp(new State(state_list));
   S->RegisterDomainMesh(Teuchos::rcp_const_cast<Mesh>(mesh));
 
-  S->RequireField("test_field", "test_field")->SetMesh(mesh)->SetGhosted(true)
-    ->SetComponent("cell", AmanziMesh::CELL, 1);
+  S->Require<CompositeVector, CompositeVectorSpace>("test_field", Tags::DEFAULT, "test_field")
```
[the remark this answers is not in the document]
lipnikov: Amanzi has DEFAULT (for current time) and and COPY (for temporary copy of default). The tag for new time is NEXT but it is not used yet, but Transport and Chemistry PK will benefit from using them. I agree that we need to standardize them.  If renaming, I prefer shorter names.  Currently, I just replaced "" with DEFAULT everywhere.

src/operators/Op_Factory.hh
```diff
@@ -0,0 +1,76 @@
+//! Helper factory for storing Ops in State
```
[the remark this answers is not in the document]
lipnikov: There are only unit tests in state/test. I plan to use them after porting mixing classes to master and adding more unit tests to operators.

src/pks/PK.hh
```diff
@@ -166,22 +167,23 @@ class PK {
   // Helper method to add a primary variable evaluator
   void AddDefaultPrimaryEvaluator_(const Key& key) {
     Teuchos::ParameterList elist;
-    elist.set<std::string>("evaluator name", key);
-    auto eval = Teuchos::rcp(new PrimaryVariableFieldEvaluator(elist));
+    elist.set<std::string>("name", key);  // probably is not used
```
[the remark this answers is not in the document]
lipnikov: I'll keep it open for a while (until we converge), since there are "names" and "tags" in Plist. If so, why not to use "name" and "tag" as well?

src/pks/PK_DomainFunctionCoupling.hh
```diff
@@ -206,7 +206,7 @@ void PK_DomainFunctionCoupling<FunctionBase>::Compute(double t0, double t1)
   Errors::Message msg;
 
   // compute reverse map: cannot move this to Init() due to lack
```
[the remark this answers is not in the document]
lipnikov: I think it was related to constructor parameters, S was not available. Amanzi will be testing coupling BCs for a student shallow water project, so it will be resolved eventually. I changed the comment to remind developers that this should be addressed.

src/pks/PK_Utils.cc
```diff
@@ -22,9 +22,9 @@ namespace Amanzi {
 void PKUtils_CalculatePermeabilityFactorInWell(
     const Teuchos::Ptr<State>& S, Teuchos::RCP<Epetra_Vector>& Kxy)
 {
-  if (!S->HasField("permeability")) return;
+  if (!S->HasData("permeability")) return;
 
-  const CompositeVector& cv = *S->GetFieldData("permeability");
+  const CompositeVector& cv = S->GetW<CompositeVector>("permeability", "permeability");
```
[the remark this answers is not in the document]
lipnikov: Thanks. Fixed it.

src/state/CMakeLists.txt
lipnikov: I do not mind. I'll do it tonight.

src/pks/eos/EOSDensityEvaluator.hh
lipnikov: A lot of work has been done on EOSs due to high P/T regime. So some care is needed. I consider this as secondary to sharing PKs.

src/pks/flow/WRMEvaluator.hh
lipnikov: Right. But not for this PR.

src/state/State.hh
lipnikov: PR passed on CI. Something is fishy. Long time ago we decided against using relatives path in Amanzi. David had a logical explanation that time. I do not want breaking this rules without his feedback.

src/mpc/test/observable_line_segment.cc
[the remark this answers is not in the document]
ecoon: In tpetra branch, we have that an Advance() call happens between two time tags, e.g. the interface is:

    PK::Advance(const Tag& old, const Tag& new);

Note, we don't _have_ to think of the tags as strongly associated with a specific time, but `State::get_time(DEFAULT)` must return something.

For an implicit PK (flow), OLD and NEW are clear, and DEFAULT==NEW is fairly obvious -- we want to vis DEFAULT, and we want to vis saturation, and saturation may only be defined at the new timestep, so that is easy enough.

For an explicit PK (transport), OLD and NEW are also clear, but DEFAULT == OLD, since intermediate steps of du/dt = f(u, x, t_old) are evaluated only at the OLD time.

When you couple them, do you assume that water content, when it is used in transport, is the OLD water content or the NEW water content?

src/pks/PK.hh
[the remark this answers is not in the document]
ecoon: PrimaryVariableFieldEvaluator evaluates 1 and only 1 field, and so does not take "names."

In the case of a secondary evaluator evaluating more than one field, "names" is necessary because the parameter list has only one string, and names is an array of strings.

    <ParameterList name="my_evaluator">
      <Parameter name="name" type="string" value="this is not used, my_evaluator instead"/>
    </ParameterList>

Allowing things to be in two places is confusing because it is unclear which takes precedence.  The ParameterList name is already required, so it made sense to remove the "name" parameter.  I don't believe the "name" parameter is even parsed at this point.

src/state/State.hh
ecoon: The other choice is to use:

    include_directory(${STATE_SOURCE_DIR})  # <-- already there
    include_directory(${STATE_SOURCE_DIR}/data)
    include_directory(${STATE_SOURCE_DIR}/evaluators)
    include_directory(${STATE_SOURCE_DIR}/io)

in every CMakeLists that uses State (most of them).

ecoon: I don't know why the PR passed on CI.  CI has no AMANZI_INSTALL_DIR that it could be pulling the old file from, does it?  Maybe the Ubuntu preprocessor is searching in subdirectories?

src/state/Tag.hh
ecoon: I think it is safe to write:

    explicit Tag(const std::string& tag) : tag_(tag) {}

By declaring it explicit, we won't get caught by accidentally passing in a string, e.g.

    S->Require<...>("mykey", "mytag");

will not compile because "mytag" cannot be converted to Tag implicitly.

src/mpc/test/observable_line_segment.cc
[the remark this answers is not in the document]
ecoon: Actually, currently on Tpetra, "" is the default, OLD time and "next" is the NEW time.  We then visualize "next".

So what that means is that, for explicit time integration, if the RHS is, e.g. `f(s(u))`, `s` is _not_ visualized.

src/state/State.hh
lipnikov: Done

src/state/State.hh
lipnikov: Done

src/pks/PK.hh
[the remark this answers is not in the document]
lipnikov: Not exactly what I mean. The logic of secondary: check "names" -> use Plist name. So, there are two ways to define name of a single secondary evaluator. Why not to use the same logic for primary: check "name" (or "names") -> use Plist name?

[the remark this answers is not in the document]
lipnikov: In addition, if primary is defined using a non-default tag, then having pair, name/tag is better then pair Plist_name/tag

src/state/State.hh
lipnikov: I have added ${STATE..}/data in a few places. I can add this include to the state directory as well. But is STATE_SOURCE_DIR is defined for this include? I can test a clean build on OSX too.

src/state/Tag.hh
lipnikov: My reasoning was that we use Key for tags as well, see derivatives. I do not mind changing it to string if you prefer... Let me know

src/state/Tag.hh
lipnikov: Yes. The purpose was to disallow implicit conversion from a string.

src/state/Tag.hh
lipnikov: Make sense from this viewpoint but what about derivatives. Then we should convert Key to string, which make no sense now. Confirm and I'll change it.

src/mpc/test/observable_line_segment.cc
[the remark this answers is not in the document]
lipnikov: We can visualize using masks. We can extend the list of visualization rules.

src/state/Tag.hh
lipnikov: So, you prefer type Tags("mytag") instead of make_tag("mytag")? I was thinking about "explicit" but decided on make_tag. I think it is just my personal style.  Any additional arguments in favor of "explicit"?

src/data_structures/CompositeVector.hh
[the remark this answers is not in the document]
lipnikov: I have no sed scripts for these changes. I used regular vi commands

src/state/Tag.hh
ecoon: done

src/pks/PK.hh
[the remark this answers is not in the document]
ecoon: So I forgot that there is no "single field" evaluator here and that secondary_variable_field_evaluator and secondary_variables_field_evaluator were merged into one.

I need to look at this logic a bit more closely anyway -- it is based on old code and misses domain set prefixes (e.g. we have to be able to infer the domain name, so that column:i-density is dependent upon column:i-pressure, for all i, without having all i in the input spec.

src/pks/eos/EOSDensityEvaluator.hh
ecoon: It looks to me that you added new EOSs but still have the original Painter et al ones for low T.  I would suggest that @dasvyat or Yu could put in EOSs for salinity, with a new evaluator, rather than the approach that was used in ATS of creating arbitrary length vectors of dependencies. 

I suggest this because it would be 10-20 fewer files to fix in ATS to use the new state.  All of my evaluators have to be updated because the Evaluate_ interface has changed.

src/state/Tag.hh
ecoon: Note to self, yes, that is the point, we are excluding this!

src/pks/PK.hh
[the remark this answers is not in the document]
lipnikov: Quite possible, that domain sets are not covered by state unit tests, so I missed them...

src/pks/eos/EOSDensityEvaluator.hh
lipnikov: What is 10 files compared to 432 in this PR ;-)

src/state/Tag.hh
lipnikov: Added explicit to have both approaches.

src/state/State.cc
```diff
@@ -871,652 +317,354 @@ State::GetModelParameters(std::string modelname) {
...
-    evaluator->second->EnsureCompatibility(Teuchos::ptr(this));
   }
 
-  // Create all data for vector fields.
-  // -- First use factories to instantiate composite vectors.
-  for (FieldFactoryMap::iterator fac_it=field_factories_.begin();
-       fac_it!=field_factories_.end(); ++fac_it) {
-    GetField_(fac_it->first)->SetData(Teuchos::rcp(new CompositeVector(*fac_it->second)));
+  for (auto& deriv : derivs_) {
+    // Some PKs allows an evalutator to be either independent or secondary.
```
ecoon: I don't understand this comment, can you expand what you mean?  This allocates memory, and should be done for all evaluator types, e.g. tpetra branch State.cc line 409.  I'm not sure why you're excluding the creation of derivatives for evaluators that are not secondary or primary?

src/state/State.hh
lipnikov: Made a clean branch on OSX 11.6. Compiled fine with clang.

src/state/State.cc
lipnikov: Independent evaluator should not have derivatives. So no memory should be allocated. But It was requested nevetheless.

lipnikov:
Regarding the first critical question about Require<CV, CVS>.  None of your examples is allowed.

src/utils/Key.hh
lipnikov: Done with tag delimiter '@'.

src/utils/Key.hh
lipnikov: Thanks, done.

src/state/State.cc
ecoon: The entity requiring the derivative should check the evaluator's IsDifferentiableWRT() method before calling RequireDerivative().  In the case of independent, this will return false, so no derivative is required.

ecoon: This is also for things like source terms, and deal with cases where a source may be a secondary eval, but not dependent upon that PK's primary variable.  For instance, consider an energy source term that is a function of pressure but not of energy (e.g. an enthalpy whose temperature is prescribed externally, but dependent on flow rate).  The energy source is a secondary variable either way, but using IsDifferentiableWRT("temperature") checks if we need dQ/dT in the jacobian of the energy residual function or not.

ecoon:
> Should we change the default value in Require(my_key, tag, owner=my_key) and make optional the last parameter in GetW(my_key, tag, owner=my_key)?

I mean, at some level, the "owner" concept was added just to have some consistency checking and force the developer to "do something extra" to get a non-const version.  One could always write:

    S->GetW<CV>("myfield", "mytag", S->GetRecord("myfield")->owner());

or something similar and not know the "password."  The fact that we have split `Get` from `GetW` does a bit of the same thing -- it forces the developer to actively think, do I really need a writeable version or not?  I didn't want the default to be we always got non-const vectors, as this is really bad for performance and logic in task parallelism, node-based parallelism, and general code correctness.

I think I'd be willing to remove the "owner" in GetW completely -- the fact that you have to write that W already is telling you to think before using it.  Setting the default would allow us to not have to refactor out the argument, and have new code not have to use it at all.  That would work for me.

src/state/State.cc
lipnikov: Done.

lipnikov:
> I think I'd be willing to remove the "owner" in GetW completely -- the fact that you have to write that W already is telling you to think before using it. Setting the default would allow us to not have to refactor out the argument, and have new code not have to use it at all. That would work for me.

How about two interfaces GetW(key, tag, owner) and GetW(key, tag) which calls GetW(key, tag, key);

src/mpc/test/observable_line_segment.cc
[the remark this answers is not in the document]
lipnikov: Closing summary: use  OLD and NEW or OLD and NEXT.  Visualization with take a tag as a global parameter.  This behavior could be modified using the existing regular expression rules.

src/pks/PK.hh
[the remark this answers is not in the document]
lipnikov: What are action items?

src/pks/eos/EOSDensityEvaluator.hh
lipnikov: @ecoon I leave this to you. I think this does not belong to this PR.

ecoon:
Ok, I have a few more changes to make to the PK interface, mostly tags into methods.  The problem is that we have one use case that makes all of this hard -- we subcycle many coupled, implicit MPCs (e.g. solve freeze-thaw on 1D columns, subcycled relative to a lateral, overland flow solve which itself a coupled flow + energy 2D surface solve and is also subcycled).

This has two implications:
1. A given PK on the column that is subcycled needs as many as four tags for e.g. water content -- the subcycle OLD and NEW, and the driver's OLD and NEW.  The driver's OLD is needed in case another PK fails even though this MPC successfully subcycled to the end of the interval.
2. A subcycling MPC can create the subcycled OLD and NEW tags, but it must give them to its children so that all quantities being subcycled together share the same tag.

The result of this is changing to:

    AdvanceStep(tag_old, tag_new);
    CommitStep(tag_old, tag_new);
    FailStep(tag_old, tag_new);

CommitStep might copy water content and pressure from new to old, while FailStep might copy it from old to new.  I think I will add these as new interface in PK, and, after we get this PR finished, we can do a second pass to agree on a common interface for PKs that can work in both Amanzi and ATS.

lipnikov:
Let PK2 be sub-cycled wrt PK1. PK2 knows that it advects fields from OLD2 to NEW2. Does "time" have these tags? Same for the other pair of tags. Finally, PK2 updates owned fields using the second set of tags. But how does it know which tags should be used for fields owned by PK1? 

In Amanzi model this solved by using copy tags and making the state filed current. So all PKs use the last available update. 
PKs know when it placed a copy to fall back in case something goes wrong.

lipnikov:
FYI, I do not think that commit step should copy new->old b/c with erases info for time derivatives which may be used for various estimates inside AdvanceStep.

ecoon:
> FYI, I do not think that commit step should copy new->old b/c this erases info for time derivatives which may be used for various estimates inside AdvanceStep.

Where do you copy new to old, and recover (on a failed step) old to new?

lipnikov:
> Where do you copy new to old, and recover (on a failed step) old to new?

Typically AdvanceStep() does both, copy and recover, see e.g. in FlowEnergyMatrixFracture_PK.  Currently, we use 
separate vectors for copy, but the plan is to use tags in an automatic way.

ecoon:
> > Where do you copy new to old, and recover (on a failed step) old to new?
> 
> Typically AdvanceStep() does both, copy and recover, see e.g. in FlowEnergyMatrixFracture_PK. Currently, we use separate vectors for copy, but the plan is to use tags in an automatic way.

What happens if you have two (weakly) coupled PKs, and the first one's Advance() is successful, but the second one fails?  Can you recover the first PK if the copy is done in Advance?

lipnikov:
CommitStep does not matter much if the code runs in the DEFAULT-tag mode which is often true for implicit solvers. For non-linear solver, NKA write directly to State filed, at least in Amanzi

lipnikov:
Each MPC will handle failure, meaning that it should create a copy of all primary fields and other quantities that could not be re-evaluated after time step reset.

lipnikov:
In your approach, we probably need to create solution vector from copies. Then, we should be careful witch tags to use in evaluators. I actually I need to write both approaches down in an abstract form to see which one is more bullet proof and does less copies.

ecoon:
Let's talk through this a bit tomorrow if you have time.  I would like to make sure that:

1. Any PK can potentially be subcycled, implicit or explicit.  In ATS we have _currently working_ use cases where we subcycle every PK and MPC.
2. We are able to subcycle any subtree of the PK tree (including subcycling MPCs of MPCs of PKs, which we do in our Arctic example).  Evaluators used in that subtree (e.g. the subgraph of the dependency graph required by PKs in the subtree) must all be valid and work on their own evaluator/Tag.
3. We support independent subcycling, e.g. an MPC that subcycles both of its children, and either one can fail or succeed.

I believe there are two possible ways to allow this:

**PK-managed copies:**

We have the following methods:

    PK::set_tags(old, new); // this is called prior to Setup(), and in most cases is called by the cycle driver with Tags::CURRENT and Tags::NEXT, the "global" old and new tags, but could be called by a subcyling MPC on its subtree with "intermediate" or "subcycling" tags.
    PK::Setup();  // given the tags, a PK requires all evaluators and data.
    PK::AdvanceStep(old, new);  // this tries to advance the step, may fail
    PK::CommitStep(old, new);  // only called if all PKs involved in the step from old to new are successful.  So this would be the entire PK tree if called by the cycle driver, or the subtree if called by a subycling MPC with tags that are unique to that subtree.  This copies fields from tag new to tag old, each PK copying their own variables.  So `Flow::CommitStep(old, new)` copies `pressure@new --> pressure@old`
    PK::FailStep(old, new);  // if any PK in the (sub) step fails, this is called immediately.  Copies fields from tag old to tag new

**Tag-wide copies:**

    PK::set_tags(old, new); // same
    PK::Setup();  //  same
    PK::AdvanceStep(old, new); //  same
    
    State::CopyTag(old, new) // for each record, if record::HasTag(old) and record::HasTag(new), copy

Instead of each PK copying in CommitStep() and FailStep(), the cycle driver (or subcycling MPC) calls `State::CopyTag()`, which loops over ALL data, and, if both old and new tags exist, copies that data.  So `State::CopyTag(old,new)` sees that both `pressure@old` and `pressure@new` have been created, and does the copy.  

A subcycling MPC can create its own "subcycling tags".  So in the case of subcycling transport, the flow + transport MPC might create tags "transport old" and "transport new", and tell transport to use these tags.  After a successful "transport subtree" (which might be transport alone, or might be transport + chemistry) step, the RT MPC calls `State::CopyTag("transport new", "transport old")`.


A couple concerns:
1. we need to make sure that things can be used outside the scope of that tag, so for instance, the advective flux in transport must use Darcy flux, which won't necessarily exist on the transport subcycling tag.
2. Vis and observations are only happening on the global Tags::NEXT (or Tags::DEFAULT, or something), so we need some logic to figure out how to mirror tags without copying.

ecoon:
> For non-linear solver, NKA writes directly into a State field, at least in Amanzi

I'm curious how this works when NKA is applied to a TreeVector that includes multiple PKs.  Clearly this wasn't possible in the old state, so I'm not sure how you used to do this.

In the new state, I hadn't considered the possibility of literally  `S::Require<TreeVector>()` which is possible now.  This might be really useful for work space that the PK really doesn't care about, like time derivatives, or "old vectors for extrapolating in time," or storing the NKA state for checkpointing, etc.

lipnikov:
Let us  list all requirement for MPC(told -> tnew) and PK(told->tnew) where told/tnew are integration times/tags.  

(a) This MPC may generate and use any tag between told and tnew.  

(b) Some field F used by MPC may be requested with different tags (interpolation of saturation), so F1(t1) and F1(t2) is possible where t1,t2 are tags/times again.

(c) Note that t1 may be outside of [told, tnew] interval in a sub-cycling PK.

(d) PK has no way to figure out tags for external fields.

I prefer your PK-managed copies approach with additional MPC to PK instructions for external fields + tags. At the moment,
we may assume that each field keeps default tag (the last tag used). Later we may pass to a sub-PK a list of external fields with tags.

lipnikov:
Amanzi populates solution vector using Tags::DEFAULT. A  copy of solution is created in AdvanceStep for recovery after failure.

ecoon:
>  with additional MPC to PK instructions for external fields + tags. 

Right, good example.  For instance, the Flow + Transport MPC could:
```
// MPCFlowTransport::Setup()
ParameterList plist("saturation");
plist.set("old tag", tag_current_);
plist.set("new tag", tag_next_);
plist.set("tag", "transport current");
auto sat_interp_eval = EvaluatorSecondaryTimeInterpolation(plist);
S->SetEvaluator("saturation", "transport current", sat_interp_eval);
```

Then, when transport needed saturation while subcycling, it would:
```
S->GetEvaluator("saturation", "transport current").Update();
S->Get<CV>("saturation", "transport current");
```
to get the interpolated value.  Note that evaluators are tag-specific, so there is nothing keeping the standard "NEXT" tag from calling van Genuchten while the "transport current" tag does interpolation.

> each field keeps default tag (the last tag used)

This is a good point -- the Record object could store it's own "default" tag, or even some simple time-based logic like:

```
void WriteVis(Vis& v, State& S, Tag tag=Tags::NEXT) {
  double time = S.get_time(tag);

  for ((name,record) in S.records) {
    if (record.HasTag(tag)) {  
      // visualize name@tag 
    } else { 
      for (other_tag in record.tags) {
        if (S.get_time(other_tag) == time) { // visualize at other_tag, break }
      }
    }
  }
}
```

ecoon:
Pushed some changes, and updated Amanzi with those changes.  The main one is to allow Copy() to not know the type -- rather than:

`S->Copy<CV>("fieldname", "source tag", "dest tag");`

the interface is now:

`S->Copy("fieldname", "dest tag", "source tag");`

The copy is done through an assignment operator that deduces the type, so there is also:

`S->GetRecord("fieldname", "dest tag")->Assign(S->GetRecord("fieldname", "source tag"));`

`Assign()` has also replaced `Set()` -- I was having a hard time remembering whether "Set" was a "shallow, pointer write" or a "deep copy."  Assign is clearly a deep copy  (in my opinion), and matches Tpetra's Assign method, which is a deep copy (remember, in Tpetra/Kokkos, operator= is a shallow, pointer copy).

It also reverse the order of "source" and "destination" tags.  This is to match Kokkos and Tpetra.

ecoon:
With this latest commit on each of Amanzi, ATS, and ats-regression-tests, the Richards tests all work in ATS.  I'm working on the above assumptions about CommitStep() and FailStep().

lipnikov:
New CommitStep() will require significant logic change in Amanzi's PK. Can just change interface and leave this logic to a separate ticket?

lipnikov:
User guide for Amanzi looks fine. Now everything is in your hands. Let me know if you need some help.

ecoon:
> New CommitStep() will require significant logic change in Amanzi's PK. Can just change interface and leave this logic to a separate ticket?

I mean, ATS will have to make some assumptions about when and who makes the copies, and those assumptions have to be generally used by all ATS PKs to deal with our current subcycling capabilities.  If I can't just do "S_next_ = S_" to back up a failed timestep, I have to do something else.  So ATS will be working on those assumptions, but since ATS and Amanzi PKs don't currently get used together, it should be a nonissue if ATS and Amanzi assume the interface means different things.

lipnikov:
Great, I want to make it less intrusive for Amanzi, but understand that ATS logic requires this. I do not want to sit on this PR for long time like, so I can modify Amanzi as soon as you commit the new interface for CommitStep().

ecoon:
@lipnikov  I just pushed a commit to master which will need to be merged into this branch.  It adds one method to the interface, HasEvaluatorList(), which is used in Observable/UnstructuredObservations (which also may need to be merged).  This shouldn't be hard, just wanted to give you a heads up.  I can do it soon.

Joe and Daniil are hopefully working on changes to ATS -- I know Joe has already begun, so hopefully this will speed things up.

lipnikov:
@ecoon The commit to master has failed.

ecoon:
It looks like this PR removes src/operators/ReconstructionCell, which is used by ATS's transport.  Please advise @lipnikov

lipnikov:
This month I'm doing refactory of Reconstruction + Limiters code, mainly under a hood and mainly limiters but interfaces are changing. Tell me which ATS branch should be adjusted. Meanwhile, changing ComputeGradient() to ComputePoly() and ReconstructionCell to ReconstructionCellGrad should work. After some cleaning, I plan to rename this function to Compute().

jbeisman:
@ecoon I just looked at 193dda65c0e6579230202ea7f26e194fdd1b3dbd and noticed that I stomped on a similar change you made to require_time a bit earlier. Sorry about that, it was unintended. Feel free to change it back.

ecoon:
@lipnikov  Ready to merge, I'll let you do the honors :D

PR #651, opened 2022-03-11
fixed bug in writing xdmf for polygonal 2d meshes

saubhagya-gatech:


ecoon:
@jd-moulton  this fixes the XDMF visualization of 3D space-dimension, 2D manifold-dimension meshes in ParaView.  The problem with visualizing surface meshes was that they were being written like polyhedrons instead of like polygons (and polygons are valid XDMF types, whereas polyhedrons are not).  

Also, this is failing tests only because of the stupid `git remote prune origin` issue again -- I forget how to make this go away so that the tests can run?  I suspect the tests will pass, but if not, it is because the tests are wrong and not because this fix is wrong.

jd-moulton:
Yes, this is failing because of the ``git remote prune origin``, which is in the Dockerfile-Amanzi and Dockerfile-ATS on the master branch.  But rather than just cherry picking that update this branch should be re-aligned with master.  It looks like it hasn't been in sync with master since late September.  So we'd need to see that merge from master come over anyway.   Is there any reason that you guys are working so far behind?  Is there an expected inconsistency or issue?  If not, one of us can go ahead and merge and push.  Should just work then.

ecoon:
Wait really?  I talked Saubhagya through the process and I'm quite certain we pulled master before unstashing the change (he had been working on 1.2) and committing.  I'll look at this again.

lipnikov:
@ecoon We used to visualize fracture meshes (2D "manifolds" in 3D) in VisIt. What is the bug then?

ecoon:
I don't know -- I would have expected trouble then too, except if your fractures were all triangles or quads.  In Saubhagya's case, any element with 5 or more faces (POLYGON, not TRI or QUAD) was not being shown, because it was being treated as a polyhedron (in which it adds an entry for each face and references all N faces into the same Element) and not a polygon (in which it simply adds one entry with N nodes).

ecoon:
> Yes, this is failing because of the `git remote prune origin`, which is in the Dockerfile-Amanzi and Dockerfile-ATS on the master branch. But rather than just cherry picking that update this branch should be re-aligned with master. 

I'm not sure what you mean by this.  If you look at the code diff relative to master, there are 3 small changes to one file.  Our process was --

1. we debugged on Amanzi release 1.2 (I didn't realize he was on 1.2)
2. git stash
3. git checkout master
4. git pull
5. git checkout -b saubhagya/mixed-element-bug
6. git stash pop
7. git commit
8. git push --set-branch-to=saubhagya/... origin/saubhagya/...

So his repo was based off an old commit, but when we branched it was correct.  And looking at the diff between this branch and master, there are no merge to be done and the only changes are the expected changes.  So I'm not sure what you mean.

lipnikov:
@ecoon This actually make sense. Most examples were for triangles and quads. If this works for non-flattened manifolds, I see no issues with this fix.

jd-moulton:
Prior to merging master this is what ``git log`` looked like

```
moulton@pn2003009 amanzi-vis % git log
commit 8d95642026781cf8abb627933dfd5cb1cde73785 (HEAD -> saubhagya/mixed-element-bug, origin/saubhagya/mixed-element-bug)
Author: saubhagya-gt <62435116+saubhagya-gt@users.noreply.github.com>
Date:   Fri Mar 11 16:07:11 2022 -0500

    typo

commit cf689297978a7d94e7e239484631c1c0bb9b70e1
Merge branch 'master' into saubhagya/mixed-element-bug
Author: saubhagya-gt <62435116+saubhagya-gt@users.noreply.github.com>
Date:   Fri Mar 11 15:48:41 2022 -0500

    fixed bug in writing xdmf for polygonal 2d meshes

commit 28b77bcc0dc84cc3c0057ec38ac718ee9c5c7dc6
Author: lipnikov <lipnikov@lanl.gov>
Date:   Wed Sep 22 19:42:30 2021 -0600

    Generalized stand-alone executable.

commit cb6dcc878c441a47e134626d855cc2854f0adb53
Author: lipnikov <lipnikov@lanl.gov>
Date:   Tue Sep 21 20:29:29 2021 -0600

    Global operator should not be built using solution's map.
```
I agree the steps you outlined make perfect sense and should have worked.   If I had to guess the ``git pull`` on master failed quietly and you didn't notice.  It would have failed for the same reason the CI was failing here, there needed to be a ``git remote prune origin``.   So I merged master in and we should be good to go now.

ecoon:
@jd-moulton your explanation makes total sense, but... I guess I'm still confused because, BEFORE you merged master in (which you can see happened after my comment), I looked at the "diff" via "Files Changed" link at the top of this page, and it showed only our changes.  Wouldn't it have shown that there was a big merge that had to happen?

Maybe not -- a "commit" is a diff, so our commit was what was shown, which was not the same thing as "git diff origin/master" which I guess is what I was expecting to have been shown...

jd-moulton:
I think the "Files Changed" is just looking at this branch/PR.  Not entirely clear how it decides what version to diff against, as PRs can have a series of changes, but probably takes the root of the branch as the starting point and diffs against that.   As you say if you had done a ``git diff origin/master`` that would have show it.  Even simpler is if look at the commit log on the PR, you see that jump back to September after Saubhagya's commits.

Issue #652, opened 2022-04-01
update cmake/SuperBuild to use find_package(Python)


ecoon:
Currently we use find_package(PythonInterp).  This has been deprecated since cmake 3.12, when find_package(Python) was introduced.  We require 3.13+ for Amanzi.

Furthermore, there is a FindPython in tools/cmake/Modules.  This looks ancient, probably is the original from 3.12.  But I don't think it is even used at this point.  

Can we pay someone to modernize all of this and just require like 3.18 or something modern-ish yet?

ecoon:
On a separate note, we require different versions of CMake for SuperBuild (3.11) and Amanzi (3.13).  These could probably be made the same.

jd-moulton:
Superseded by #700, which drops back to find\_package(PythonInterp) for now.  Also CMake version issue is addressed in #701.

PR #653, opened 2022-04-01
passes python executable provided in bootstrap to a few TPLs

ecoon:
This just passes PYTHON_EXECUTABLE, provided via --with-python argument to bootstrap, into Trilinos and SEACAS.

refs #652  but does not try to fix that.


ecoon:
@jd-moulton  Is this superfluous now?  Maybe it has been fixed on konstantin/state @lipnikov

lipnikov:
@jd-moulton The development branch could get pieces of this accidentally, not the complete PR.

PR #655, opened 2022-04-08
Subgrid Source Sink Fixed

saubhagya-gatech:
The final source and sink from Domain functions "FirstOrderExchange.hh" and "SubgridReturn.hh" were missing factors of molar density liquid, porosity and ponded depth (saturation).

### DIFF
--- a/src/pks/PK_DomainFunctionSubgridReturn.hh
+++ b/src/pks/PK_DomainFunctionSubgridReturn.hh
@@ -53,7 +53,12 @@ class PK_DomainFunctionSubgridReturn : public FunctionBase,
 
  private:
 
-  std::string field_out_suffix_, copy_field_out_key_, dset_;
+  std::string field_out_suffix_, copy_field_out_key_, dset_; 
+  std::string saturation_key_, saturation_copy_;
+  std::string porosity_key_;
+  std::string porosity_copy_;
+  std::string molar_density_key_;
+  std::string molar_density_copy_;
 };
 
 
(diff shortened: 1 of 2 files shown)


lipnikov:
@ecoon Okay. Do you approve this fix?

src/pks/PK_DomainFunctionSubgridReturn.hh
```diff
@@ -76,6 +81,19 @@ void PK_DomainFunctionSubgridReturn<FunctionBase>::Init(
   
   copy_field_out_key_ = blist.get<std::string>("copy subgrid field", "default");
 
+ // can "surface" prefix for this field be read automatically?? it is hardcoded now, or may be not?? 
+  saturation_key_= Keys::readKey(blist, blist.get<std::string>("domain name", "surface"), 
+                          "saturation liquid", "ponded_depth");
```
ecoon, line 86: This is a little weird, even if it is correct for our use case.  I'm hesitant to make the default so non-obvious, however.

src/pks/PK_DomainFunctionSubgridReturn.hh
```diff
@@ -76,6 +81,19 @@ void PK_DomainFunctionSubgridReturn<FunctionBase>::Init(
   
   copy_field_out_key_ = blist.get<std::string>("copy subgrid field", "default");
 
+ // can "surface" prefix for this field be read automatically?? it is hardcoded now, or may be not??
```
ecoon, line 84: I think what you're doing now is right.  Since the destination's domain is not "surface", this can only be read from the input file.

ecoon (approved):
Approved, but this will need to be merged with konstantin/state.

lipnikov:
I cherry-picked this commit and upgraded it to new State. Will push soon.

ecoon:
Closing this now then as it has been merged into konstantin/state -- it will be merged when that branch is merged.

PR #666, opened 2022-06-10
Changes to konstantin/state needed to get ATS tests passing

ecoon:
Significant changes:

- addition of aliasing capabilities: Records and Evaluators can now be "aliased" by which we mean two tags, each pointing to the same {data, evaluator}.  
- `State::RequireEvaluator(key, Tag::NEXT)` will now first search for an existing evaluator with a tag that contains the word "next" in it, and point to that if it exists before trying to create a new evaluator.  This allows for cases where, e.g. flow is subcycled, so "saturation" exists on "flow subcycled next" tag.  We don't want a new evaluator, we just want to get the subcycled next (assuming the times are the same).  No checks are currently made to ensure these are at the same time.
- Please use Keys::readTag() to parse a tag from a parameter list rather than directly reading a string.  This will allow the default to be set correctly in the case where we wish to change the default tag from "" to something more descriptive later on.
- Chemistry PK now uses tags.  These tags may need to get cleaned up a bit.
- Amanzi may need to call set_tags() if it wishes to subcycle.  See, for instance, ATS source code `src/pks/mpc/mpc_subcycled.hh`
- Vis currently ONLY writes variables at the DEFAULT tag.  This could be changed if needed?
- Checkpoint, on the other hand, reads and writes variables at ALL tags.  This is important to save "old" variables for exact checkpoint/restart, which should be possible now once the solution history is put into state!
 
EDIT: this is WIP.  All "shared" unit tests work, as do ATS regression tests, but a few of Amanzi's unit tests fail still for me:

	180 - chemistry_benchmarks_1d (Failed)
	209 - mpc_flow_reactive_transport (Failed)
	220 - mpc_benchmark_single (Failed)

and I have not checked benchmarks/Amanzi regression tests/user guide tests.  I will not get to these this week as I'm on vacation, so if @lipnikov have some time, please feel free!

### DIFF
--- a/src/state/IO.cc
+++ b/src/state/IO.cc
@@ -49,7 +49,7 @@ void WriteVis(Visualization& vis, const State& S)
         // visualized. However, since overwriting of attributes does not work
         // properly, we skip them.  This should get fixed by writing attributes
         // as an attribute of the step or something similar. FIXME --ETC
-        if (!r->second->ValidType<double>() && !r->second->ValidType<int>()) {
+        if ((!r->second->ValidType<double>()) && (!r->second->ValidType<int>())) {
           // Should we vis all tags or just the default tag?
           // -- write all tags
           // r->WriteVis(vis, nullptr);
--- a/src/state/data/DataFactory.hh
+++ b/src/state/data/DataFactory.hh
@@ -42,7 +42,10 @@ class DataFactory {
 
   DataFactory(DataFactory&& other) noexcept : p_(std::move(other.p_)) {}
 
-  void swap(DataFactory& other) noexcept { p_.swap(other.p_); }
+  void swap(DataFactory& other) noexcept {
+    p_.swap(other.p_);
+    data_p_.swap(other.data_p_);
+  }
 
   DataFactory& operator=(DataFactory other) {
     if (&other != this)
@@ -52,12 +55,12 @@ class DataFactory {
 
   bool HasType() const { return p_.get(); }
 
-  template <typename T, typename F> bool ValidType() const {
-    return p_->ValidType<T, F>();
-  }
-
   template <typename T> bool ValidType() const {
     return data_p_->ValidType<T>();
+  }
+
+  template <typename T, typename F> bool ValidType() const {
+    return p_->ValidType<T,F>();
   }
 
   template <typename T, typename F> const F& Get() const {
--- a/src/state/data/RecordSet.hh
+++ b/src/state/data/RecordSet.hh
@@ -170,12 +170,12 @@ class RecordSet {
   }
 
   template <typename T, typename F>
-  bool ValidType() {
+  bool ValidType() const {
     return factory_.ValidType<T,F>();
   }
 
   template <typename T>
-  bool ValidType() {
+  bool ValidType() const {
     return factory_.ValidType<T>();
   }
 
--- a/src/state/test/state_container.cc
+++ b/src/state/test/state_container.cc
@@ -23,6 +23,12 @@ TEST(STATE_CREATION) {
   State s;
   s.Require<double>("my_double", Tags::DEFAULT, "my_double");
   s.Setup();
+
+  CHECK(s.GetRecord("my_double", Tags::DEFAULT).ValidType<double>());
+  CHECK(!s.GetRecord("my_double", Tags::DEFAULT).ValidType<int>());
+
+  CHECK(s.GetRecordSet("my_double").ValidType<double>());
+  CHECK(!s.GetRecordSet("my_double").ValidType<int>());
 }
 
 TEST(STATE_ASSIGNMENT) {
--- a/src/state/test/state_data_factory.cc
+++ b/src/state/test/state_data_factory.cc
@@ -7,7 +7,15 @@ using namespace Amanzi;
 
 TEST(NULL_FACTORY) {
   DataFactory fac = dataFactory<double, NullFactory>();
+  CHECK(fac.ValidType<double>());
+  CHECK(!fac.ValidType<int>());
+  CHECK(!fac.ValidType<Vec>());
+
   auto s = fac.Create();
+  CHECK(fac.ValidType<double>());
+  CHECK(!fac.ValidType<int>());
+  CHECK(!fac.ValidType<Vec>());
+
   s.Assign(1.1);
   CHECK_EQUAL(1.1, s.Get<double>());
 }
@@ -25,9 +33,22 @@ TEST(VEC_FACTORY) {
   g_constructor_calls_main = 0;
   g_constructor_calls_copy = 0;
   DataFactory fac = dataFactory<Vec, VecFactory>();
+
+  bool valid = fac.ValidType<Vec>(); CHECK(valid);
+  valid = fac.ValidType<Vec,VecFactory>(); CHECK(valid);
+  valid = !fac.ValidType<double>(); CHECK(valid);
+  valid = !fac.ValidType<Vec,double>(); CHECK(valid);
+  valid = !fac.ValidType<double,VecFactory>(); CHECK(valid);
+
   fac.GetW<Vec, VecFactory>().set_size(2);
 
   auto s = fac.Create();
+  valid = fac.ValidType<Vec>(); CHECK(valid);
+  valid = fac.ValidType<Vec,VecFactory>(); CHECK(valid);
+  valid = !fac.ValidType<double>(); CHECK(valid);
+  valid = !fac.ValidType<Vec,double>(); CHECK(valid);
+  valid = !fac.ValidType<double,VecFactory>(); CHECK(valid);
+
   s.GetW<Vec>().v[0] = 1.1;
   CHECK_EQUAL(1.1, s.Get<Vec>().v[0]);
   CHECK_EQUAL(1, g_constructor_calls_main);


src/output/HDF5_MPI.cc
```diff
@@ -705,7 +705,8 @@ void HDF5_MPI::createTimestep(double time, int iteration, const std::string& tag
     Teuchos::XMLObject tmp("Xdmf");
     tmp.addChild(addXdmfHeaderLocal_("Mesh",time,iteration));
     std::stringstream filename;
-    filename << H5DataFilename() << "." << iteration << tag << ".xmf";
+    //filename << H5DataFilename() << "." << iteration << tag << ".xmf";
```
ecoon, line 708: clean up dead code

src/pks/PK_DomainFunctionFirstOrderExchange.hh
```diff
@@ -148,6 +148,7 @@ void PK_DomainFunctionFirstOrderExchange<FunctionBase>::Compute(double t0, doubl
         val_vec[i] = -(*(*uspec)->first->second)(args)[i] * tcc[i][*c] * ws_[i][*c] * phi_[i][*c] * mol_dens_[i][*c];
       }
       value_[*c] = val_vec;
+      if (*c == 0) std::cout << "Computed FOE src term with val = " << val_vec[0] << std::endl;
```
ecoon, line 151: remove debugging cruft

src/state/evaluators/EvaluatorIndependent.cc
```diff
@@ -20,7 +20,7 @@ namespace Amanzi {
 // ---------------------------------------------------------------------------
 EvaluatorIndependent_::EvaluatorIndependent_(Teuchos::ParameterList& plist)
     : my_key_(Keys::cleanPListName(plist.name())),
-      my_tag_(make_tag(plist.get<std::string>("tag", ""))),
+      my_tag_(plist.get<std::string>("tag", Tags::DEFAULT.get())),
```
ecoon, line 23: call readTag()

src/state/evaluators/EvaluatorPrimary.cc
```diff
@@ -23,7 +23,7 @@ namespace Amanzi {
 // ---------------------------------------------------------------------------
 EvaluatorPrimary_::EvaluatorPrimary_(Teuchos::ParameterList& plist)
     : my_key_(Keys::cleanPListName(plist.name())),
-      my_tag_(make_tag(plist.get<std::string>("tag", ""))),
+      my_tag_(plist.get<std::string>("tag", Tags::DEFAULT.get())),
```
ecoon, line 26: call readTag()

ecoon (commented):
few minor changes

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -123,9 +109,6 @@ Alquimia_PK::Alquimia_PK(Teuchos::ParameterList& pk_tree,
 
   chem_engine_->GetAqueousKineticNames(aqueous_kinetics_names_);
   number_aqueous_kinetics_ = aqueous_kinetics_names_.size();
-  
-  // verbosity object
-  vo_ = Teuchos::rcp(new VerboseObject("Alquimia_PK:" + domain_, *cp_list_));
```
lipnikov, line 128: I would keep this hard-coded name. When Amanzi have multiple time periods, the auto-generated name for the list may not be what we want to report in logs for clarity.  On the other hand, if this name follows certain rules (it is in fact based on Key), then we can remove non-informative prefix. But will be hard to agree on such rules.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -484,12 +476,12 @@ void Alquimia_PK::XMLParameters()
...
-                                 AlquimiaState& state,
-                                 AlquimiaAuxiliaryData& aux_data)
+        AlquimiaProperties& mat_props,
+        AlquimiaState& state,
+        AlquimiaAuxiliaryData& aux_data,
+        const Tag& sat_tag)
 {
-  auto tcc = S_->GetW<CompositeVector>(tcc_key_, Tags::DEFAULT, passwd_).ViewComponent("cell", true);
-  CopyToAlquimia(cell, tcc, mat_props, state, aux_data);
+  CopyToAlquimia(cell, aqueous_components_, mat_props, state, aux_data, sat_tag);
```
lipnikov, line 484: I do not remember why tcc was used here, b/c of gas components, b/c of tags, or b/c of something else?

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -570,16 +563,21 @@ void Alquimia_PK::CopyToAlquimia(int cell,
     }
   }
 
-  const auto& water_saturation = *S_->Get<CompositeVector>(saturation_key_).ViewComponent("cell", true);
+  // These should be DEFAULT (or really tag_next_) but are left as CURRENT to
+  // mimic a bug in the old code for testing.  THIS SHOULD NOT BE COMMITTED and
+  // MUST be fixed prior to MERGE WITH MASTER! See ATS #129
```
lipnikov, line 568: Is it addressed?

src/pks/chemistry/Amanzi_PK.cc
```diff
@@ -112,17 +97,6 @@ Amanzi_PK::Amanzi_PK(Teuchos::ParameterList& pk_tree,
   number_aqueous_components_ = comp_names_.size();
   number_free_ion_ = number_aqueous_components_;
   number_total_sorbed_ = number_aqueous_components_;
-
-  // verbosity object
-  vo_ = Teuchos::rcp(new VerboseObject("Amanzi_PK:" + domain_, *cp_list_));
```
lipnikov, line 117: Same comment as for Alquimia PK.

src/pks/chemistry/Chemistry_PK.cc
```diff
@@ -164,56 +187,57 @@ void Chemistry_PK::Setup()
...
+  // Aqueous species
   if (number_aqueous_components_ > 0) {
     if (!S_->GetRecordW(tcc_key_, passwd_).initialized()) {
-      InitializeCVField(S_, *vo_, tcc_key_, Tags::DEFAULT, passwd_, 0.0);
+      InitializeCVField(S_, *vo_, tcc_key_, tag_next_, passwd_, 0.0);
     }
- 
-    InitializeCVField(S_, *vo_, primary_activity_coeff_key_, Tags::DEFAULT, passwd_, 1.0);    
-    InitializeCVField(S_, *vo_, free_ion_species_key_, Tags::DEFAULT, passwd_, 1.0e-9);    
+    set_aqueous_components(S_->GetPtrW<CompositeVector>(tcc_key_, tag_next_, passwd_)
```
lipnikov, line 195: originally, set_aqueous_components() was used in reactive transport MPC PK  and other PKs as a replacement of the current tag-based design. We probably need to run chemistry benchmarks to make sure that failed chemistry 
step does not destroy tcc provided by the transport.

src/state/State.hh
```diff
@@ -122,8 +122,8 @@ class State {
   typedef std::map<Key, Teuchos::RCP<AmanziMesh::DomainSet>> DomainSetMap;
   typedef std::map<Key, Teuchos::RCP<Functions::MeshPartition> > MeshPartitionMap;
 
-  using RecordSetMap = std::unordered_map<Key, std::unique_ptr<RecordSet> >;
-  using EvaluatorMap = std::unordered_map<Key, std::unordered_map<Tag, Teuchos::RCP<Evaluator> > >;
+  using RecordSetMap = std::map<Key, std::unique_ptr<RecordSet> >;
```
lipnikov, line 125: I'm curious, is it for performance or for code uniformity?

lipnikov (approved):
In general looks quite clean. Major concerns: (1) CI fails during compile time; (2) treatment of tcc and aqueous_components in reactive transport MPC PKs. Let me know if you need me to look at (1) and chemistry benchmark problems.

lipnikov:
FYI: mpc_flow_reactive_transport is crashing in Chemistry_PK::Setup()->RequireEvaluator()->CreateEvaluator()-validateEntryExists(). Some new is needed in XML...

lipnikov (commented):
Chemistry is failing...

src/pks/chemistry/Alquimia_PK.cc
ecoon: Rather than that, would it be ok to set the PK::name_ attribute with this value?  

Typically my expectation is that this is the "user's" name, not the "developer's" name, but if you want to hard-code the name, then it should probably set the attribute so that all uses of name_ are consistent (not just the VO printing).

src/pks/chemistry/Alquimia_PK.cc
ecoon: No idea.  It was the only place that this was done (e.g. getting tcc from state rather than using the local copy), so I assumed it was oversight/inconsistency.

src/pks/chemistry/Alquimia_PK.cc
ecoon: Yes, I'll remove the comment.

src/pks/chemistry/Chemistry_PK.cc
ecoon: Haha, yeah, that would be bad.  I have not run all the Amanzi tests yet - just shared ATS/Amanzi unit tests and ATS regression tests.  I started that process but it is not done (see my edits to description to see status) and I have not run the user guide/regression tests.

I would just as soon remove this local copy, but that seemed an "after merge" thing to do.

src/state/State.hh
ecoon: It is for correctness.  Consider a case:

Rank 0:  "a", "b", "c"
Rank 1:  "a", "c"
Rank 2: "b", "c"

"c" is on MPI_COMM_WORLD, while "a" and "b" are on sub-communicators.  By keeping these in an ordered map, we can guarantee that there will be no "deadlock" cases when we iterate over the keys for things that do collective operations (e.g. vis).

src/pks/chemistry/Alquimia_PK.cc
lipnikov: PK::name_ sounds like a reasonable compromise. VO output should be more rigid in my opinion. This helps debugging user's problem.

lipnikov:
Since transport + chemistry cannot be a shared test for now, one way to proceed is to add two tests to interoperability directory. Do u have a good XML with subcycling ATS transport and Amanzi chemistry?

src/pks/chemistry/Alquimia_PK.cc
ecoon: Please see #672 .  I will revert this change for now but think we need to come up with a standard and use it across the code.

src/pks/chemistry/Chemistry_PK.cc
ecoon: chemistry_benchmarks_1d passes now -- does that resolve your concerns?

ecoon:
@lipnikov  I believe that this will now pass CI, and can be merged once it does so.  Once that is done, I believe that konstantin/state can be merged into master.

PR #668, opened 2022-06-13
added update to cybernetic model to pflotran patch

ghammond86:
This patch enables inhibition of the cybernetic rate due to depletion of NH4+.


ghammond86:
@smolins @saubhagya-gatech 

Can either of you review the merge request? It should be straightforward. I recommend using the "View File" option as the diff produces odd results.

jd-moulton (commented):
Patch applies cleanly in my rebuild of TPLs, and all of our tests are passing.  Glenn confirmed this particular functionality is tested within PFLOTRAN's tests as well.  I will update TPL version on the master branch so users are aware that the TPLs have changed and they would need to rebuild to get this update in PFLOTRAN.

PR #677, opened 2022-07-07
cleanup to geometry based on work in mesh_refactor

ecoon:
- makes enums class enums
- renames Region methods according to newer Amanzi standards
- removes the Domain concept, which does not appear to be used anywhere


ecoon:
@lipnikov  @jd-moulton  I haven't tested this on structured yet, so may still fail.  Most of this is more maintenance in prep for mesh refactor, but the one concept change is removing the Domain object.  It doesn't appear to be used anywhere -- do you use it for anything at this point?

src/utils/RaggedArray.hh
```diff
@@ -0,0 +1,277 @@
+/*
```
lipnikov, line 1: Looks like this file has no test. Is it a placeholder for something?

src/executables/AmanziUnstructuredGridSimulationDriver.cc
```diff
@@ -89,39 +88,19 @@ AmanziUnstructuredGridSimulationDriver::Run(const Amanzi::Comm_ptr_type& comm,
...
-  // Create the simulation domain
-  std::unique_ptr<Amanzi::AmanziGeometry::Domain> sim_domain;
+  //------------ GEOMETRIC MODEL ----------------------------
   Teuchos::RCP<Amanzi::AmanziGeometry::GeometricModel> geom_model;
   auto geom_timer = Teuchos::TimeMonitor::getNewCounter("Geometric Model creation");
 
   { // context for timer
     Teuchos::TimeMonitor geom_tm(*geom_timer);
 
-    Teuchos::ParameterList domain_params = plist_->sublist("domain");
```
lipnikov, line 101: This list was introduced by Rao to specify dimensionality of the model. I see no other way to get it. May be "geometric model"  would be a better name.

ecoon: Whups, right, I compiled this, then cleaned up some more comments, and removed this line by mistake.  For now I'll just revert this bit, keeping the "domain" sublist.

For what it is worth, I don't think we actually use this number anywhere.  In ATS at least, this is hard-coded as 3, and used with 3 even on 2D runs.  The GM doesn't actually respect or use the spatial dimension anywhere I don't think -- it certainly can store regions that are both 2D regions and 3D regions in the same GM, and we use the same GM on both the surface (2D) and subsurface (3D) meshes in the same run.

I suspect that removing the GM's spatial dimension completely won't change anything...

ecoon: Ok, there are two places where the GM's dimension is used:

1. If a Mesh_MSTK is constructed from file, then the GM's dimension is compared to the mesh's space dimension and an error is thrown if they don't match.  No such check is done on meshes that are generated.

https://github.com/amanzi/amanzi/blob/master/src/mesh/mesh_mstk/Mesh_MSTK.cc#L218

2. In AmanziUnstructuredGridSimulationDriver, the mesh->space_dimension() is compared to the gm->dimension() and an error is thrown if they don't match.  This at least checks both case 1 and the generated case.

https://github.com/amanzi/amanzi/blob/master/src/executables/AmanziUnstructuredGridSimulationDriver.cc#L269



In both cases, all this checks is that the user wrote the right number in the xml.  I'm not sure that this is a useful thing to keep.  I think there used to be some more checks in place on regions to make sure that dimensioned regions (e.g. RegionBox) had the same dimension as the GM, but those got removed when we started using the same GM for flattened, derived meshes.  Unless we really want to rethink GM (e.g. making multiple GMs), I'd suggest we just remove the use of the GM's dimension completely.  Alternatively, just write, at the same L269 of AmanziUnstructuredGridSimulationDriver:

`geom_model->set_dimension(mesh->space_dimension());`

and remove the parameter from the input spec.

ecoon: Note, the useful checks are when the region is used with a mesh to get entities.  When that actually happens, we DO check that the mesh's space_dimension matches the region's dimension.  That check is important and currently produces useful error messages for the user to fix their region.

src/utils/RaggedArray.hh
ecoon: Ok, I removed this -- there is a test, but that test relies on having DenseVector and WhetStone::Tensor classes written for Kokkos.

Julien pointed out that this won't actually be used in the Mesh refactor anyway -- we'll use a simpler data structure -- so no need to put it in now.

ecoon:
For now, I'll leave the "domain" parameter list as is, even if there is no `Domain` object.  This can be changed if we come to some consensus on a way forward and/or refactor GMs.

Assuming tests pass, I'll merge this later today.

ecoon:
@lipnikov  Sorry, didn't mean to flag you there -- I accidentally clicked the "rerequest review" button and don't seem to be able to cancel the request.

PR #678, opened 2022-07-24
Treatment of Python executable variable

lipnikov:
If the variable PYTHON_EXECUTABLE is already set before the call (as a normal or cached variable) then the search will not occur inside find_program(). When bootstrap.sh defines it as the empty string, it is cached and not updated inside FindPythonInterp module. We need to unset it or change the logic of bootstrap. This PR removes empty string from cache.


jd-moulton (commented):
CMake lists PythonInterp as deprecated since version 3.12.   So we should probably update to the recommended Python3 or Python.   The catch is that I quickly tried Python3 and it didn't work, and Python worked, but then it could happily find a python 2 interpreter - leaving us to check for that.  It would be great if you could get Python3 working.

lipnikov:
I think PythonInterp will be supported until CMake 4 is released. Are you unhappy with it current behavior?

jd-moulton:
I'm happy with the current behavior, it solves the problem with our bootstrap script.  I was just a little concerned about adding technical debt.  But I think the question of the python interpreter that we find and pass on is going to need updates as the TPLs change their handling of python anyway.  So let's go with this for now.

jd-moulton (approved):

ecoon:
Is this consistent with the TPLs change?  Is there a reason this is still open or can we merge it @jd-moulton  @lipnikov

PR #679, opened 2022-07-25
fixes bugs in DomainSet and vis introduced in state merge

ecoon:
Also adds an abbreviate() method to Keys, which relies on a global dictionary of abbreviations to shorten PK names or variable names.

I think @lipnikov  uses a similar simple substitution somewhere in State -- it may make sense to merge this concepts.


ecoon:
Note this also changes Vis files to remove the ".cell.0" suffix on all variables.  Cell quantities are the only thing we visualize, so we don't need to include that, and the ".0" is unnecessary when there is only one vector.  

This will likely annoy some people (especially postprocessing scripts) but I think it is past time to fix this obnoxious artifact in the vis output...

PR #682, opened 2022-08-15
Naren/sw

nvohra0016:
Flow over dry beds. Scheme as in [Beljadid et al' 16].


src/pks/shallow_water/NumericalFlux_Rusanov.cc
```diff
@@ -55,7 +55,7 @@ std::vector<double> NumericalFlux_Rusanov::Compute(
   SL = std::fabs(uL) + std::sqrt(g_*hL);
   SR = std::fabs(uR) + std::sqrt(g_*hR);
 
-  Smax = std::max(SL, SR);
+  Smax = 1.e0 * std::max(SL, SR);
```
lipnikov, line 58: Multiplication by 1.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -49,7 +49,8 @@ ShallowWater_PK::ShallowWater_PK(Teuchos::ParameterList& pk_tree,
   if (found != std::string::npos) pk_name.erase(0, found + 2);
 
   // Create miscellaneous lists.
-  Teuchos::RCP<Teuchos::ParameterList> pk_list = Teuchos::sublist(glist, "PKs", true);
+  Teuchos::RCP<Teuchos::ParameterList> pk_list =
```
lipnikov, line 52: I think we decided to allow 100 characters in one line. Please change this in the script and rerun it.

lipnikov (changes requested):
I suggest first to reformat for 100-character lines as in master::src/.clang-format

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -265,37 +277,40 @@ void ShallowWater_PK::Initialize()
...
   // compute B_c from B_n for well balanced scheme (Beljadid et. al. 2016)
   S_->Get<CV_t>(bathymetry_key_).ScatterMasterToGhosted("node");
 
+  cell_area_max_ = 0.0;
   for (int c = 0; c < ncells_owned; ++c) {
-    const Amanzi::AmanziGeometry::Point &xc = mesh_->cell_centroid(c);
-        
+    const Amanzi::AmanziGeometry::Point& xc = mesh_->cell_centroid(c);
+
+    cell_area_max_ = std::max(cell_area_max_, mesh_->cell_volume(c) * mesh_->cell_volume(c));
```
lipnikov, line 299: Do a * a outside the loop. Also, rename cell_area_max to cell_area2_max_ to code readibility.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -428,25 +442,23 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
 
   Epetra_MultiVector& h_old = *soln_->SubVector(0)->Data()->ViewComponent("cell");
   Epetra_MultiVector& q_old = *soln_->SubVector(1)->Data()->ViewComponent("cell");
-  
-  for (int c = 0; c < ncells_owned; ++c) {
-    double factor = inverse_with_tolerance(h_old[0][c]);
+
+  for (int c = 0; c < ncells_wghost; ++c) {
```
lipnikov, line 446: h_old has no ghost values. Even if you require them, you need to distribute them.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -477,26 +489,30 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
   }
-  
+
   Epetra_MultiVector& h_temp = *soln_->SubVector(0)->Data()->ViewComponent("cell");
   Epetra_MultiVector& q_temp = *soln_->SubVector(1)->Data()->ViewComponent("cell");
 
   // update solution
-  for (int c = 0; c < ncells_owned; ++c) {
-    double factor = inverse_with_tolerance(h_temp[0][c]);
+  for (int c = 0; c < ncells_wghost; ++c) {
```
lipnikov, line 514: Similar to above, h_temp has no ghost values.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +555,136 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+//--------------------------------------------------------------
+// Total Depth ht = h + B (Evaluate value at edge midpoint for a polygonal cell)
+// Reconstruct if necessary for positivity
+//--------------------------------------------------------------
+double
+ShallowWater_PK::TotalDepthEdgeValue(int c, int e)
+{
+  double ht_edge; // value to return
+
+  auto& ht_c = *S_->GetW<CompositeVector>(total_depth_key_, passwd_).ViewComponent("cell", true);
```
lipnikov, line 577: Can you use total_depth_key_ here?  It is called inside a time integrator with a vector not necessary from the State.

lipnikov (changes requested):
My understanding that requested work could be essential. Please reply before I continue.

src/pks/shallow_water/test/shallow_water_dry_bed.cc
```diff
@@ -0,0 +1,313 @@
...
+  S->InitializeFields();
+  S->InitializeEvaluators();
+  S->set_time(0.0);
+  SWPK.Initialize();
+
+  dry_bed_setIC(mesh, S, icase);
+  S->CheckAllFieldsInitialized();
+
+  const auto& B =
+    *S->Get<CompositeVector>("surface-bathymetry").ViewComponent("cell");
```
lipnikov, line 209: Looks like this file was not reformatted for 100-char lines

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +558,136 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+  cell_is_partially_wet = false;
+  cell_is_dry = false;
+  cell_is_fully_flooded = false;
+
+  // calculate maximum bathymetry value on cell nodes
+  double Bmax = 0.0;
+  for (int i = 0; i < cnodes.size(); ++i) { Bmax = std::max(B_n[0][cnodes[i]], Bmax); }
+
+  // characterize cell based on [Beljadid et al.' 16]
+  if ((ht_c[0][c] > Bmax) && (ht_c[0][c] - B_c[0][c] > 1.e-15)) {
```
lipnikov, line 602: This is dangerous. If h and B about 100 m, the difference has only 14 valid digits. Should we use a relative tolerance here? Also, it would be better to place it at the top of routine. Another good place to various tolerances is ShallowWaterDefs.hh as we do in other PKs.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +558,136 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+  bool cell_is_dry, cell_is_fully_flooded, cell_is_partially_wet;
+  cell_is_partially_wet = false;
+  cell_is_dry = false;
+  cell_is_fully_flooded = false;
+
+  // calculate maximum bathymetry value on cell nodes
+  double Bmax = 0.0;
+  for (int i = 0; i < cnodes.size(); ++i) { Bmax = std::max(B_n[0][cnodes[i]], Bmax); }
+
+  // characterize cell based on [Beljadid et al.' 16]
```
lipnikov, line 601: Add journal also.

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: So would it be better to pass ht_c as an argument in the function?

lipnikov: Eventually, this may be required. But I'm more concern of the logic, say, in context of RK2.

nvohra0016: Hmm, I see, but we recompute the total depth ht_c in the time integrator (ShallowWater_TI.cc), so it should be fine? 
[
](https://github.com/amanzi/amanzi/blob/ed9f9057854faa9c84bcd91b46d12fab7a0f5829/src/pks/shallow_water/ShallowWater_TI.cc#L55)

That means that the intermediate values of ht_c are used, say, in RK2 for the total depth reconstruction?

nvohra0016: I have made some changes in the implementation. Now only the nodal bathymetry is passed as an argument and the rest of the values (like total depth) are passed as scalars.

lipnikov: Okay.

src/pks/shallow_water/ShallowWater_PK.hh
```diff
@@ -81,32 +81,41 @@ class ShallowWater_PK : public PK_Physical,
...
   std::vector<double> PhysicalFlux_x(const std::vector<double>&);
 
-  std::vector<double> NumericalFlux_x(std::vector<double>&, std::vector<double>&);
+  std::vector<double>
+  NumericalFlux_x(std::vector<double>&, std::vector<double>&);
   std::vector<double> NumericalFlux_x_Rusanov(const std::vector<double>&, const std::vector<double>&);
   std::vector<double> NumericalFlux_x_CentralUpwind(const std::vector<double>&, const std::vector<double>&);
 
-  std::vector<double> NumericalSource(const std::vector<double>&, int);
+  std::vector<double> NumericalSource(int c, double htc, double Bc, const Epetra_MultiVector B_n);
```
lipnikov, line 104: You need to pass by reference.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +557,130 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+
+  bool cell_is_dry, cell_is_fully_flooded, cell_is_partially_wet;
+  cell_is_partially_wet = false;
+  cell_is_dry = false;
+  cell_is_fully_flooded = false;
+
+  // calculate maximum bathymetry value on cell nodes
+  double Bmax = 0.0;
+  for (int i = 0; i < cnodes.size(); ++i) { 
+  	Bmax = std::max(B_n[0][cnodes[i]], Bmax);
```
lipnikov, line 594: From one angle it makes sense to precompute and cache this field since bathymetry is static. However, it may vary in the future due deposition, so probably we should not do it. Just a comment to think a lit, no action is needed.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +557,130 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+  if ((htc > Bmax) && (htc - Bc > 0.0)) {
+    cell_is_fully_flooded = true;
+  } else if (std::abs(htc- Bc) < 1.e-15) {
+    cell_is_dry = true;
+  } else {
+    cell_is_partially_wet = true;
+  }
+
+  // depth poisitivity based on [Beljadid et al.' 2016, Computers and Fluids]
+  if (cell_is_fully_flooded == true) {
```
lipnikov, line 607: We use if (a) {} in amanzi, if a is boolean.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +557,130 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+
+      mesh_->node_get_coordinates(face_nodes[0], &x0);
+      mesh_->node_get_coordinates(face_nodes[1], &x1);
+
+      Amanzi::AmanziGeometry::Point tria_edge0, tria_edge1;
+
+      tria_edge0 = xc - x0;
+      tria_edge1 = xc - x1;
+
+      Amanzi::AmanziGeometry::Point area_cross_product = (0.5) * tria_edge0 ^ tria_edge1;
```
lipnikov, line 635: = 0.5 * ((xc - x0) ^ (xc - x1))  would be shorter

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +557,130 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+
+      double area = norm(area_cross_product);
```
lipnikov, line 637: I think this is even better  area = norm((xc - x0) ^ (xc - x1)) / 2;

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +557,130 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+    double mu_eps_sum = 0.0;
+
+    for (int f = 0; f < cfaces.size(); ++f) {
+      Amanzi::AmanziGeometry::Point x0, x1;
+      int edge = cfaces[f];
+
+      Amanzi::AmanziMesh::Entity_ID_List face_nodes;
+      mesh_->face_get_nodes(edge, &face_nodes);
+
+      mesh_->node_get_coordinates(face_nodes[0], &x0);
```
lipnikov, line 627: Suggest to define n0 = face_nodes[0] and use it later, for code readability.

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -537,25 +557,130 @@ bool ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
...
+      tria_edge0 = xc - x0;
+      tria_edge1 = xc - x1;
+
+      Amanzi::AmanziGeometry::Point area_cross_product = (0.5) * tria_edge0 ^ tria_edge1;
+
+      double area = norm(area_cross_product);
+
+      double epsilon;
+
+      double ht_rec_x0 = htc;
```
lipnikov, line 641: What is the reason for renaming htc?  Compiler will optimize it out... still?

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -565,53 +690,52 @@ std::vector<double> ShallowWater_PK::NumericalSource(
...
-  int ncells_owned = mesh_->num_entities(AmanziMesh::CELL, AmanziMesh::Parallel_type::OWNED);
+  S_->GetEvaluator(discharge_key_).Update(*S_, passwd_);
+  auto& q_c = *S_->GetW<CV_t>(discharge_key_, Tags::DEFAULT, discharge_key_).ViewComponent("cell", true);
+
+  int ncells_wghost = mesh_->num_entities(AmanziMesh::CELL, AmanziMesh::Parallel_type::ALL);
   AmanziMesh::Entity_ID_List cfaces;
-  
-  for (int c = 0; c < ncells_owned; c++) {
+
+  for (int c = 0; c < ncells_wghost; c++) {
```
lipnikov, line 738: Why do u need ghosts here?

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -645,11 +774,12 @@ double ShallowWater_PK::get_dt()
     Teuchos::OSTab tab = vo_->getOSTab();
     *vo_->os() << "switching from reduced to regular cfl=" << cfl_ << std::endl;
   }
-  
-  if (iters_ < max_iters_)
+
+  if (iters_ < max_iters_) {
```
lipnikov, line 778: Are brackets {} due to re-formatting?

src/pks/shallow_water/test/shallow_water_1D_dry_bed.cc
```diff
@@ -0,0 +1,289 @@
...
+      mesh->node_get_coordinates(face_nodes[1], &x1);
+
+      Amanzi::AmanziGeometry::Point tria_edge0, tria_edge1;
+
+      tria_edge0 = xc - x0;
+      tria_edge1 = xc - x1;
+
+      Amanzi::AmanziGeometry::Point area_cross_product = (0.5) * tria_edge0^tria_edge1;
+
+      double area = norm(area_cross_product);
```
lipnikov, line 106: I suggest area = norm ((xc - x0)^(xc - x1)) / 2;

src/pks/shallow_water/test/shallow_water_1D_dry_bed.cc
```diff
@@ -0,0 +1,289 @@
...
+  
+  while ((t_new < 5.0) && (iter >= 0)) {
+   
+    double t_out = t_new;
+            
+    if (iter % 10 == 0) {
+               
+      io.InitializeCycle(t_out, iter, "");
+                
+      io.WriteVector(*hh(0), "depth", AmanziMesh::CELL);
```
lipnikov, line 240: Eventually, we may create help I/O functions and use them in all tests, probably after a merge.

src/pks/shallow_water/test/shallow_water_lake_at_rest.cc
```diff
@@ -1,6 +1,6 @@
 /*
   Shallow water PK
- 
+
```
lipnikov, line 3: If this file was not modified due to new capabilties, let us remove all re-formatting by checking out the original version. Reason is that future potential debugging will be easier with smaller number of changes.

lipnikov (approved):
Approval is contingent to resolving my comments.

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: Fixed. Only needed positivity of (ht_c - B_c). Had 1.e-15 for debugging purposes earlier.

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: Fixed. Was using that for debugging purposes earlier.

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: I see what you mean.

src/pks/shallow_water/test/shallow_water_lake_at_rest.cc
nvohra0016: Okay

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: I am unsure. They may have moved due to re-formatting.

PR #692, opened 2022-10-06
fixes some logic on communication

ecoon:
fixes some logic about communication, removes some unnecessary communication, and asserts that communication can happen when it is requested

- ASSERTS when Scatter is called but vector !ghosted
- PDE_DiffusionFV had unnecessary/unused communication
- adds some missing communication that previously relied on the user to have already done

Note that this may mean that Amanzi can remove some Scatter() calls?

### DIFF
--- a/src/data_structures/CompositeVector.cc
+++ b/src/data_structures/CompositeVector.cc
@@ -339,6 +339,7 @@ void
 CompositeVector::ScatterMasterToGhosted(std::string name, bool force) const {
   // NOTE: allowing const is a hack to allow non-owning PKs to nonetheless
   // update ghost cells, which may be necessary for their discretization
+  AMANZI_ASSERT(ghosted_);
 #ifdef HAVE_MPI
 #if MANAGED_COMMUNICATION
   if (ghosted_ && ((!ghost_are_current_[Index_(name)]) || force)) {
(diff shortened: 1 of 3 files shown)


ecoon:
@lipnikov  Can you take a look at this branch?  There aren't too many changes.  Possible reasons for the breaking tests:

1. I removed a "scatter" call from faces of dkr/dp for PDE_DiffusionFV that I believe is not required by the PDE_DiffusionFV code.  There was a scatter there, but ghost entries were never used, so the scatter should have been unnecessary *for that part of the code.*  The scatter *could* have been necessary for *later* parts of the code (e.g. the PK, not the PDE_DiffusionFV operator), in which case the scatter should be called where it is needed.
2. There is a new AMANZI_ASSERT added that ensures that if you call Scatter on a CompositeVector that is not Ghosted(), it fails.  Previously, the Scatter call on a non-ghosted vector silently did nothing.  I believe you should never call Scatter on a non-ghosted vector, so making this an ASSERT likely exposed cases where the CV was not Ghosted but was being used as if it was ghosted.

I'm guessing the problem is 2, as the tests are serial and 1 should only be possible in parallel tests.

Can you check that this is so, and add the needed SetGhosted() call to whichever PK is requiring that vector?

ecoon:
Note that the reason I found this was because derivatives do not get the "ghosted" flag from their parent vector.  So just because kr is ghosted does not mean that "dkr/dp" is ghosted.  See amanzi/ats#161 to see what I had to fix in ATS -- not sure if this is the same problem in Amanzi's PKs or not.

ecoon:
Ok, looks like some changes to the logic with regards to CompositeVector::Update() seems to allow the "ghosted" flag to carry over properly for Amanzi.  @lipnikov I believe this should fix a bug in parallel Amanzi multiphase runs.

PR #697, opened 2022-10-18
Sw positivity

nvohra0016:
Updated the algorithm for positivity correction for _fully flooded cells_ by scaling the gradient accordingly. The new implementation ensures mass conservation.

### DIFF
--- a/src/pks/shallow_water/ShallowWater_PK.cc
+++ b/src/pks/shallow_water/ShallowWater_PK.cc
@@ -347,6 +347,9 @@ ShallowWater_PK::Initialize()
 
     S_->GetRecordW(total_depth_key_, Tags::DEFAULT, passwd_).set_initialized();
   }
+  
+  // initialize cell flag (for positivity)
+  cell_flag_.resize(ncells_owned, 0);
 
   InitializeCVField(S_, *vo_, velocity_key_, Tags::DEFAULT, passwd_, 0.0);
   InitializeCVField(S_, *vo_, discharge_key_, Tags::DEFAULT, discharge_key_, 0.0);
@@ -459,6 +462,9 @@ ShallowWater_PK::AdvanceStep(double t_old, double t_new, bool reinit)
     vel_c[0][c] = factor * q_old[0][c];
     vel_c[1][c] = factor * q_old[1][c];
     ht_c[0][c] = h_c[0][c] + B_c[0][c];
+    
+    // cell flags for depth positivity
+    cell_flag_[c] = 0;
   }
 
   // update source (external) terms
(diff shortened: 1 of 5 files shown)


src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -615,10 +615,28 @@ ShallowWater_PK::TotalDepthEdgeValue(
...
+    if (cell_visited_[c] == false) {
+      Amanzi::AmanziMesh::Entity_ID_List cfaces;
+      mesh_->cell_get_faces(c, &cfaces);
+      
+      double alpha = 1.0; // gradient scaling factor
+      for (int f = 0; f < cfaces.size(); ++f) {
+        const auto& xf_tmp = mesh_->face_centroid(cfaces[f]);
+        ht_edge = total_depth_grad_->getValue(c, xf_tmp);
+        if (ht_edge - BathymetryEdgeValue(cfaces[f], B_n) < 0.0) {
+          alpha = std::min(alpha, 0.9 * (BathymetryEdgeValue(cfaces[f], B_n) - htc) / (ht_edge - htc));
```
lipnikov, line 627: What is the role of 0.9?

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -615,10 +615,28 @@ ShallowWater_PK::TotalDepthEdgeValue(
...
+        }
+      }
+      ht_grad[0][c] = alpha * ht_grad[0][c];
```
lipnikov, line 630: Why not C-style code?

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -615,10 +615,28 @@ ShallowWater_PK::TotalDepthEdgeValue(
 
   // depth poisitivity based on [Beljadid et al.' 2016, Computers and Fluids]
   if (cell_is_fully_flooded) {
-    ht_edge = total_depth_grad_->getValue(c, xf);
-    if (ht_edge - BathymetryEdgeValue(e, B_n) < 0.0) { 
-      ht_edge = htc; 
+    if (cell_visited_[c] == false) {
```
lipnikov, line 618: Can we make this additional limiting a separate function and call it right after the gradient calculation of ht?

src/pks/shallow_water/ShallowWater_PK.cc
```diff
@@ -615,10 +615,28 @@ ShallowWater_PK::TotalDepthEdgeValue(
...
+      Amanzi::AmanziMesh::Entity_ID_List cfaces;
+      mesh_->cell_get_faces(c, &cfaces);
+      
+      double alpha = 1.0; // gradient scaling factor
+      for (int f = 0; f < cfaces.size(); ++f) {
+        const auto& xf_tmp = mesh_->face_centroid(cfaces[f]);
+        ht_edge = total_depth_grad_->getValue(c, xf_tmp);
+        if (ht_edge - BathymetryEdgeValue(cfaces[f], B_n) < 0.0) {
```
lipnikov, line 626: Since edge value is used so often, may be we should make it a "face" component of the static field B?

This is not a suggestion for this PR, but for the next step. For similar reasons, Bmax could be precomputed as well.

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: As a "safety" factor to get alpha less than the critical value that ensures corrected total depth = 0.

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: Thank you, fixed.

src/pks/shallow_water/ShallowWater_PK.cc
nvohra0016: Changed the code. Please see.

PR #698, opened 2022-10-18
Optional parameter for initialization logic

lipnikov:
Initialization is enforced in one place only, when we read initial conditions in State. Amanzi "initialize" rule allows to overwrite data using state->initial condition list.

### DIFF
--- a/src/state/State.cc
+++ b/src/state/State.cc
@@ -715,7 +715,7 @@ void State::InitializeFields(const Tag& tag)
           flag = "[ok]";
           Teuchos::ParameterList sublist = state_plist_.sublist("initial conditions").sublist(e.first);
           sublist.set<double>("time", t_ini);
-          e.second->Initialize(sublist);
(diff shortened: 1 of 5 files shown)


src/state/State.cc
```diff
@@ -715,7 +715,7 @@ void State::InitializeFields(const Tag& tag)
           flag = "[ok]";
           Teuchos::ParameterList sublist = state_plist_.sublist("initial conditions").sublist(e.first);
```
ecoon, line 716: This shouldn't be necessary, but does it make sense to add a conditional here?

```
if state_plist_.sublist("initial conditions").isSublist(e.first) ...
```

rather than always constructing the (potentially empty sublist) and forcing initialize.  It _shouldn't_ do anything with an empty list, but might as well skip any potential problems?

ecoon (approved):
This should be fine and shouldn't affect ATS.

PR #728, opened 2022-12-08
refactor of Mesh library for use with Kokkos

JulienLoiseau:
- Add the cache support discussed and worked on at https://github.com/amanzi/mesh_cache_prototype2
- This MR is for review purpose only


src/mesh/MeshCache_impl.hh
```diff
@@ -184,258 +204,1044 @@ MeshCache<MEM>::getNumEntities(const Entity_kind kind, const Parallel_type ptype
...
+  return nodes[i];
+}
+
+template<MemSpace_type MEM>
+template<typename ViewType> // Can be normal, unmanaged or vector 
+KOKKOS_INLINE_FUNCTION
+void
+MeshCache<MEM>::getCellNodes(const Entity_ID c, ViewType& nodes) const{
+  
+  nodes = RaggedGetter<MEM,AccessPattern::DEFAULT>::get(data_.cell_nodes_cached,
```
JulienLoiseau, line 757: In this version, due to the order of the getCellNodes in the framework we are caching the values or using the Framework to compute. 
We could: 
- Ignore the order and change the tests to just check that all values are here 
- Find a way to reorder the nodes following ExodusII order.

src/mesh/MeshDefs.hh
```diff
@@ -38,12 +38,6 @@ using Entity_GID = int;
 using Set_ID = int;
 using size_type = Kokkos::View<int*, Kokkos::DefaultHostExecutionSpace>::size_type;
 
-enum class MemSpace_type {
```
JulienLoiseau, line 41: The name of the enum needs to be changed. 
What is the latest convention we agreed on? `_type` or `_kind`?

src/mesh/MeshSets.hh
```diff
@@ -24,14 +24,12 @@
 #include "RegionLabeledSet.hh"
 #include "RegionLogical.hh"
 #include "MeshDefs.hh"
+//#include "Mesh_Algorithms.hh"
```
JulienLoiseau, line 27: Cleanup comments

src/mesh/Mesh_Helpers.hh
```diff
@@ -9,5 +9,7 @@
 */
```
JulienLoiseau, line 9: Should this file name be changed to MeshHelpers.hh (and MeshHelpers_decl + MeshHelpers_impl) to follow the naming convention?

src/mesh/Mesh_Helpers_impl.hh
```diff
@@ -413,6 +414,19 @@ Entity_ID_List getCellFaceAdjacentCells(const Mesh_type& mesh,
 }
 
 
+template<class Mesh_type> 
+KOKKOS_INLINE_FUNCTION
+AmanziGeometry::Point getFaceCentroid(const Mesh_type& mesh, const Entity_ID f)
```
JulienLoiseau, line 419: These need to be replaced by the function provided by Konstantin

src/mesh/mesh_extracted/MeshExtractedManifold.cc
```diff
@@ -35,21 +35,24 @@ namespace AmanziMesh {
 * Light-weigthed constructor
 ****************************************************************** */
 MeshExtractedManifold::MeshExtractedManifold(
-    const Teuchos::RCP<const Mesh>& parent_mesh,
+    const Teuchos::RCP<MeshFramework>& parent_mesh,
```
JulienLoiseau, line 38: Note: The MeshExtractedManifold is passed a MeshFramework for initialization

src/mesh/mesh_extracted/MeshExtractedManifold.cc
```diff
@@ -35,21 +35,24 @@ namespace AmanziMesh {
...
+    const Teuchos::RCP<Teuchos::ParameterList>& plist,
     bool request_faces, bool request_edges,
     bool flattened)
-  : Mesh(comm, gm, plist, request_faces, request_edges),
+  : MeshFramework(comm, gm, plist),
     parent_mesh_(parent_mesh),
     flattened_(flattened)
 {
+  
+  mesh_cache_ = Teuchos::rcp(new MeshCache<MemSpace_type::HOST>(parent_mesh, plist));
```
JulienLoiseau, line 51: Note: A temporary mesh_cache_ is created to be used later in the MeshExtractedManifold

src/mesh/mesh_extracted/test/mesh_extracted_fracture.cc
```diff
@@ -72,19 +73,25 @@ void RunTest(const std::string regname, int* cells, int* edges) {
 
     // extract fractures mesh
     try {
-      RCP<Mesh> mesh = Teuchos::rcp(new MeshExtractedManifold(mesh3D, setname, AmanziMesh::FACE,
+      RCP<MeshFramework> mesh = Teuchos::rcp(new MeshExtractedManifold(mesh3D, setname, AmanziMesh::Entity_kind::FACE,
```
JulienLoiseau, line 76: Look carefully at these changes, this test is not passing yet. 
A MeshExtractedManifold if a MeshFramework, a MeshCache is then created from it below and used in the audit, which only work on MeshCache (Mesh).

src/mesh/mesh_extracted/test/mesh_surface_flattened.cc
```diff
@@ -41,16 +41,22 @@ TEST(MESH_SURFACE_FLATTENED) {
...
-  RCP<Mesh> mesh = Teuchos::rcp(new MeshExtractedManifold(mesh3D, "Top side", AmanziMesh::FACE,
+  RCP<MeshFramework> mesh = Teuchos::rcp(new MeshExtractedManifold(mesh3D, "Top side", AmanziMesh::Entity_kind::FACE,
                                                           comm, gm, plist, true, false, flatten));
-
-  int ncells = mesh->cell_map(false).NumGlobalElements();
-  int nfaces = mesh->face_map(false).NumGlobalElements();
-  int nnodes = mesh->node_map(false).NumGlobalElements();
-  int mfaces = mesh->exterior_face_map(false).NumGlobalElements();
+  
+  RCP<Mesh> mesh_cached = Teuchos::rcp(new Mesh(mesh));
```
JulienLoiseau, line 50: Same as other test, a mesh_cache is created. 
This test doesnt run with: 
```
/home/jloiseau/AMANZI/amanzi_mesh/src/mesh/mesh_extracted/test/mesh_surface_flattened.cc:29:1: error: Failure in MESH_SURFACE_FLATTENED: Unhandled exception: Could not retrieve mesh entities of kind=3 for set "Top side"
```

src/mesh/mesh_factory/test/mesh_factory_extract_comms.cc
```diff
@@ -68,37 +68,38 @@ TEST (EXTRACT_COMM_SPLIT) {
...
     // END SEGMENT
 
-    auto newmesh = meshfactory.create(parent_mesh, setnames, Amanzi::AmanziMesh::FACE, flatten);
-    if (local_count == 0) {
-      CHECK(newmesh == Teuchos::null);
-    } else {
-      CHECK(newmesh != Teuchos::null);
-    }
+    // Which constructor? 
+    //auto newmesh = meshfactory.create(parent_mesh, setnames, Amanzi::AmanziMesh::Entity_kind::FACE, flatten);
```
JulienLoiseau, line 97: A problem choosing the constructor here since parent_mesh is a MeshCache and this factory take MeshFramrwork. Maybe just use `->getMeshFramework()`?

src/mesh/test/mesh_fracture_extraction.cc
```diff
@@ -18,6 +18,7 @@
 
 #include "AmanziComm.hh"
 #include "RegionFactory.hh"
+#include "Mesh.hh"
```
JulienLoiseau, line 21: Basically the same problem of factory with using MeshFramework when it requires MeshCache or vice-versa. 
See other tests: extraction_fraction and surface_flattened

src/mesh/test/mesh_surface_extraction.cc
```diff
@@ -197,10 +196,10 @@ TEST(MESH_SURFACE_EXTRACTION_GENERATED_EXTRACTED_MANIFOLD)
     auto mesh = fac.create(parent_mesh, {"Top Face Plane"}, AmanziMesh::Entity_kind::FACE, true);
 
     // make sure we got what we think we got
-    {
-      auto mesh_as_manifold = Teuchos::rcp_dynamic_cast<MeshExtractedManifold>(mesh);
-      CHECK(mesh_as_manifold != Teuchos::null);
-    }
+    //{
+    //  auto mesh_as_manifold = Teuchos::rcp_dynamic_cast<MeshExtractedManifold>(mesh);
```
JulienLoiseau, line 200: The mesh is of type MeshCache when the MeshExtractedManifold is derived from MeshFramework.

JulienLoiseau:
@ecoon I added a bunch of comments/questions on the MR. 
I will still be looking on my side and also work on changing my branch to be master with only mesh being changed, so I can work on the sed scripts.

JulienLoiseau:
All tests in mesh are passing except the ones involving the MeshExtractedManifold, the comments I added point to these problems.

src/mesh/Mesh_Helpers.hh
ecoon: If they help the MeshCache, it should be MeshCache_Helpers.

If they help the MeshFramework, it should be MeshFramework_Helpers.

Otherwise it should probably be MeshHelpers.

src/mesh/mesh_extracted/MeshExtractedManifold.cc
ecoon: This is fine -- all meshes are constructed at startup and so no frameworks will have been deleted yet (I think).

src/mesh/mesh_extracted/MeshExtractedManifold.cc
JulienLoiseau: SHould just accept a Framework, avoid to create a MeshCache inside.

ecoon: Should accept a MeshCache now, previous comment is wrong.

lipnikov:
30% and more of tests are failing

src/data_structures/CompositeVector.hh
```diff
@@ -329,7 +329,7 @@ class CompositeVector {
   int PutScalar(const std::string& name, double scalar);
 
   // v(name,i,:) = scalar[i]
-  int PutScalar(const std::string& name, std::vector<double> scalar);
+  int PutScalar(std::string name, AmanziMesh::Double_List scalar);
```
lipnikov, line 332: why name is not a const string&

src/data_structures/MatrixFE.cc
```diff
@@ -90,7 +90,7 @@ MatrixFE::SumIntoMyValues(const int* row_indices,
                           const Epetra_SerialDenseMatrix& vals)
 {
   int ierr(0);
-  std::vector<double> row_vals(vals.N());
+  AmanziMesh::Double_List row_vals(vals.N());
```
lipnikov, line 93: name "Double_List" seems far away from a vector<double>

src/data_structures/SuperMap.hh
```diff
@@ -37,9 +37,9 @@ class CompositeVector;
 class TreeVectorSpace;
 class TreeVector;
 
-namespace AmanziMesh {
-class Mesh;
-}
+//namespace AmanziMesh {
+//class Mesh;
+//}
```
lipnikov, line 43: Remove these lines.

src/data_structures/test/data_structures_matrix_fe_speed.cc
```diff
@@ -383,8 +383,8 @@ TEST(FE_MATRIX_NEAREST_NEIGHBOR_TPFA_MatrixFE)
 //   Teuchos::RCP<GraphFE> graph =
```
lipnikov, line 383: Too many commented out lines. Suggest to add "FIXME"

src/executables/VerifyHex.cc
```diff
@@ -106,11 +106,11 @@ dump_output(const int& me, Amanzi::AmanziMesh::Mesh& mesh, const std::string& fi
   std::fill(mypart.begin(), mypart.end(), -1);
 
   // Amanzi::AmanziMesh::Set_ID_List setids;
```
lipnikov, line 108: Now clear why this block was commented out. Anyway, this is optional.

src/functions/FunctionBilinearAndTime.hh
```diff
@@ -12,11 +12,11 @@
 
 A piecewise bilinear function that is additionally linear interpolated in time.
 
-Define :math:`i(x) = i : x_i < x <= x_{{i+1}}` and similarly :math:`j(y) = j : y_j < y <= y_{{j+1}}` 
+Define :math:`i(x) = i : x_i < x <= x_{{i+1}}` and similarly :math:`j(y) = j : y_j < y <= y_{{j+1}}`
 for monotonically increasing :math:`x_i` and :math:`y_j`.
 
 Given a two-dimensional array :math:`u_{{i,j}}`, :math:`f` is then defined by
-bilinear interpolation on
```
lipnikov, line 19: Strange to see tailing spaces, since we re-formatted master many times.

src/mesh/MeshCache_impl.hh
```diff
@@ -0,0 +1,1762 @@
...
+      auto normal = data_.face_normals.get<MEM>(f, 0);
+      auto nat_dir = data_.face_normal_orientations.get<MEM>(f, 0);
+      if (abs(nat_dir) == 1) {
+        return nat_dir * normal;
+      } else {
+        auto fcells = getFaceCells(f, Parallel_kind::ALL);
+        assert(fcells.size() == 2);
+
+        // average normals oriented from lower to higher GIDs
+        int pos_i =
```
lipnikov, line 453: I would rename pos_i to pos to squeeze into one line

src/mesh/MeshDefs.hh
```diff
@@ -6,115 +6,195 @@
...
+using cDouble_View = View_type<const double>;
+
+//
+// View for host only
+//
+using Entity_ID_List = std::vector<Entity_ID>;
+using Entity_GID_List = std::vector<Entity_ID>;
+using Entity_Direction_List = std::vector<int>;
+using Point_List = std::vector<AmanziGeometry::Point>;
+using Double_List = std::vector<double>;
```
lipnikov, line 82: Such type of renaming makes thing unnecessary complex. Any reasons for doing it?

src/mesh/mesh_extracted/test/mesh_surface_flattened.cc
```diff
@@ -64,19 +70,19 @@ TEST(MESH_SURFACE_FLATTENED)
 
   AmanziGeometry::Point p;
   for (int n = 0; n < nnodes; ++n) {
-    mesh->node_get_coordinates(n, &p);
+    p = mesh_cached->getNodeCoordinate(n);
```
lipnikov, line 73: Should this be  auto& p = mesh_cached->getNodeCoordinates(n);

src/mesh/mesh_factory/test/mesh_factory.cc
```diff
@@ -14,7 +14,7 @@
 
 #include "dbc.hh"
 #include "GeometricModel.hh"
-#include "../MeshFactory.hh"
+#include "../MeshFrameworkFactory.hh"
```
lipnikov, line 17: We typically avoid relative path in include statements.

src/mesh/test/mesh_geometry.cc
```diff
@@ -152,11 +157,11 @@ TEST(MESH_GEOMETRY_3CUBE_EXO)
 
 //   for (const auto& frm : frameworks) {
 //     std::cout << std::endl
-//               << "Testing 3D Box 3x3x3 Exo with " << AmanziMesh::framework_names.at(frm) << std::endl
```
lipnikov, line 155: Duplicated code?

src/mesh/test/test_mesh_maps.cc
```diff
@@ -0,0 +1,555 @@
...
+      // Check the normal with respect to each connected cell
+
+      Amanzi::AmanziMesh::Entity_ID_View cellids;
+      mesh->getFaceCells(i, Amanzi::AmanziMesh::Parallel_kind::ALL, cellids);
+
+      for (int k = 0; k < cellids.size(); k++) {
+        int dir;
+        Amanzi::AmanziGeometry::Point normal_wrt_cell = mesh->getFaceNormal(i, cellids[k], &dir);
+
+        //            Amanzi::AmanziMesh::Entity_ID_View cellfaces;
```
lipnikov, line 122: Are these lines still needed?

src/mesh_functions/test/boundary_function.cc
```diff
@@ -134,29 +137,25 @@ TEST_FIXTURE(reference_mesh, values1)
...
+  face_list = mesh->getSetEntities(BACK, FACE, Parallel_kind::ALL);
+  for (auto f = face_list.begin(); f != face_list.end(); ++f) CHECK_EQUAL(3.0, bf.find(*f)->second);
 }
 
 
 TEST_FIXTURE(reference_mesh, values2)
 {
   // Create the function f(t,x,y,z) = t * (x + 2y + 3z)
-  std::vector<double> c(1, 1.0);
+  AmanziMesh::Double_List c(1, 1.0);
```
lipnikov, line 154: std::vector<double> is much more readable

src/mpc/CycleDriver.cc
```diff
@@ -313,7 +313,7 @@ CycleDriver::ReportMemory()
     double global_ncells(0.0);
     double local_ncells(0.0);
     for (State::mesh_iterator mesh = S_->mesh_begin(); mesh != S_->mesh_end(); ++mesh) {
-      Epetra_Map cell_map = (mesh->second.first)->cell_map(false);
+      Epetra_Map cell_map = (mesh->second.first)->getMap(AmanziMesh::Entity_kind::CELL, false);
```
lipnikov, line 316: Should we avoid a copy constructor here?

src/mpc/ObservableAqueous.hh
```diff
@@ -37,11 +37,11 @@ class ObservableAqueous : public virtual Observable {
   virtual int ComputeRegionSize();
 
  protected:
-  double CalculateWaterTable_(State& S, AmanziMesh::Entity_ID_List& ids);
+  double CalculateWaterTable_(State& S, AmanziMesh::Entity_ID_View& ids);
 
   int obs_boundary_;
   bool obs_planar_;
-  std::vector<double> vofs_;
+  AmanziMesh::Double_View vofs_;
```
lipnikov, line 44: I do not see how Double_View is more readable then std::vector<double>?

src/mpc/ObservableSolute.hh
```diff
@@ -43,7 +43,7 @@ class ObservableSolute : public virtual Observable {
 
  protected:
   bool obs_boundary_, obs_planar_;
-  std::vector<double> vofs_;
+  AmanziMesh::Double_View vofs_;
```
lipnikov, line 46: Same as before, I prefer std::vector<double> unless there is a strong argument against?

src/operators/LimiterCellDG.cc
```diff
@@ -212,10 +210,10 @@ LimiterCellDG::LimiterHierarchicalDG_(const WhetStone::DG_Modal& dg,
 
     for (int i = 0; i < nfaces; i++) {
       int f = faces[i];
-      mesh_->face_get_nodes(f, &nodes);
+      auto nodes = mesh_->getFaceNodes(f);
 
-      mesh_->node_get_coordinates(nodes[0], &x1);
-      mesh_->node_get_coordinates(nodes[1], &x2);
+      x1 = mesh_->getNodeCoordinate(nodes[0]);
```
lipnikov, line 215: const auto& x1 = mesh_-> ... is probably better. If so, place make a ticket to clean these later

lipnikov (approved):
I made a few simple comments. A short response would be enough.

src/operators/LimiterCellDG.cc
ecoon: Yes, that's right -- the "shared memory across the loop iterations" both isn't necessary now and is harmful when we go to Kokkos, which parallelizes these loop bodies across threads (and so you can't share memory across the loop iterations).

#774

src/mpc/ObservableAqueous.hh
ecoon: `using Double_View = Kokkos::MeshView<T*, typename Impl::MemorySpace<MEM>::space>;`

(MeshDefs.hh)

It's not a std::vector anymore.  You can make it a std::vector if needed:

```
auto [entity_ids, vofs_view] = mesh->getSetEntitiesAndVolumeFractions(...);
std::vector<double> vofs_vector = AmanziMesh::asVector(vofs_view);
```

src/mpc/ObservableSolute.hh
ecoon: see above

src/mpc/CycleDriver.cc
ecoon: yes, fixing

src/mesh_functions/test/boundary_function.cc
ecoon: So we already had (on master)

`using Entity_ID_List = std::vector<Entity_ID>;`

And we added, e.g.

`using Entity_ID_View = Kokkos::View<Entity_ID*, ...>;`

So to parallel those, we needed:

`using Double_View = Kokkos::View<double*, ...>;`. which gets used for any mesh-cached doubles (e.g. volume of fluids)

I'm not sure if we need Double_List though, unless maybe in the internals of MeshFramework when talking about volume of fluids, but certainly not when passing arguments to Functions.

So this should definitely be std::vector<double>.  The FunctionPolynomial constructor accepts std::vector<double>.  I don't know how many places std::vector<double> was renamed Double_List, but hopefully not many places!

I'll fix this one now and do a quick grep to see how much this shows up.

ecoon: Ok, it's not a ton outside of Mesh.  I think it is ok to use it in like mesh_logical -- there we accept Double_Lists and then convert them to Double_Views, which are the thing we store in the Cache.  But it shouldn't be used outside of the mesh library.

I think we can just sed them back to std::vector outside of src/mesh.

src/mpc/ObservableAqueous.hh
lipnikov: My fault. All such comments should be related to Double_List not Double_View.

src/mesh_functions/test/boundary_function.cc
lipnikov: Let us make a ticket and do this cleaning later

src/mesh/test/test_mesh_maps.cc
ecoon: this is dead code -- removed the file

src/mesh/test/mesh_geometry.cc
ecoon: I'm not sure -- this reads a .par file and not a .exo file.  That mostly works, but the test fails because of a bug in MSTK that has not yet been fixed.  That bug is only a problem for prepartitioned exo meshes with node sets, which we rarely use (I think, it's been a while).

So I left the test there for now in the hope that maybe MSTK gets fixed eventually and then we can turn the test back on?

src/mesh/mesh_extracted/test/mesh_surface_flattened.cc
ecoon: Probably actually just 

`const auto p = mesh_cached->getNodeCoordinates(n);`

Creating new points is cheap since they are on the stack (Point::d is a double[3], no allocation needed) so they are treated as "plain old data."

I not sure if `const auto& p` will work, but it would be better if it did work.

At the end of the day, we'll need to remove all these once we start using Kokkos/threading over loops anyway.  This is another case of #774

src/mesh/MeshDefs.hh
ecoon: see below, happy to discuss more, but Entity_ID_List is all over the code.

I think eventually we'll want to make this be std::span<Entity_GID>, not std::vector<>, so when that happens, we may appreciate the aliasing.

src/mesh/MeshCache_impl.hh
ecoon: Oh yuck, I don't like how clang-formatted that one at all...

ecoon: Ok, but even if pos_i is renamed to a single character, the line is still > 100 chars, so would get split.

I still don't like where clang-format split the lines there, but I'm not sure what is better or how to convince clang-format to do it.

src/functions/FunctionBilinearAndTime.hh
ecoon: Agreed.

I wonder if it is because it is in comment blocks?  Maybe clang-format changed something here?  My clang-format version may have changed on my new laptop/OSX.

src/executables/VerifyHex.cc
ecoon: It was commented out 11 years ago... I think it's safe to remove?  Removing these big block comments, and some Rao-style empty lines.

src/data_structures/test/data_structures_matrix_fe_speed.cc
ecoon: removing them -- they are redundant and covered by the subsequent tests.  I think I left them in to support debugging performance, but they've been commented out for 5+ years...

src/data_structures/MatrixFE.cc
ecoon: see below

src/mesh/test/mesh_surface_extraction.cc
```diff
@@ -266,10 +265,10 @@ TEST(MESH_SURFACE_EXTRACTION_EXO_EXTRACTED_MANIFOLD)
...
```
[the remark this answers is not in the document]
ecoon: oh good catch.  I thought I fixed all of these, but apparently not.  There are a few in BlockVector as well, fixing as many as I can find now.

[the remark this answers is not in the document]
ecoon: (these are merging errors, they were fixed in master, then unfixed and merged and not caught)

src/mesh/MeshDefs.hh
ecoon: Note, I think these are only really intended for use in the mesh library, where we've been changing the data structures regularly! Outside of the mesh library one should only get View objects.

src/mesh/mesh_factory/test/mesh_factory.cc
ecoon: Yeah, I don't think it should be needed either, because the directory is already there from `include_directory()` in CMakeLists.txt.

jd-moulton (approved):
Builds and tests passing for me on OSX.  CI is passing too, so I think we are good to go.

PR #733, opened 2023-01-26
Change std::vector to View

JulienLoiseau:
- Add MeshView and MeshDualView helpers for Amanzi. 
- MeshCache and MeshFramework is based on Kokkos::Views (through MeshView and MeshDualView).


ecoon:
merging this now, no one is "in between" these branches, and @lipnikov  said he could look at this tonight.

PR #735, opened 2023-02-20
netcdf-fortran cmake fixes

fmyuan:
When using CMake to build NetCDF-Fortran, there appears 2 unknown libraries called 'MPI::MPI_Fortran' and 'netcdff_c', which actually could not be found anywhere.


jd-moulton:
In order to collect a few small updates and keep the TPL collection version number consistent and avoid forcing people to rebuild too often, I created a new branch and changed the base of the PR.

jd-moulton (approved):
This doesn't seem to show up on all systems, but cleaning it up seems like a good idea.

PR #736, opened 2023-02-20
Fix on confusing versions between AMANZI_TPLS and AMANZI.

fmyuan:
In AMANZI_TPLS version file, it's better to rename AMANZI to AMANZI_TPLS, because it's actually TPLS version rather than AMANZI version.


jd-moulton:
Changed base for PR to include with collection of other TPL updates.

ecoon:
@jd-moulton  This makes sense to me, can we merge this in now as a part of the upgrade to 0.98.7?

jd-moulton:
Agree we need to pick this up, it got stuck in the older 0.98.7 update path, and that was ultimately superseded by the Trilinos and Hypre update in the actual jump to 0.98.7.  

I think my plan is going to be to rename this TPLs branch to cover 0.98.8, and separate this PR from that so it can move quickly.

jd-moulton (approved):
This change does make the naming more consistent to intent/purpose is clearer.

jd-moulton:
Improves consistency of internal variable names in CMakeLists.txt for the TPLs.

PR #757, opened 2023-07-12
Update to SubsetOf

lipnikov:
One use case of this change in a multistage simulation: 
pressure (cell + face) on stage 1 
pressure (cell + boundary_face) on stage 2.


ecoon (approved):
sure, that makes perfect sense to me: "boundary_face" is a subset of "face"

PR #758, opened 2023-07-17
better definition of IsDifferentiableWRT

ecoon:
This change of IsDifferentiableWRT allows evaluators to have the key as a dependency, but NOT be differentiable with respect to that dependency.  This allows other evaluators to overload this method and turn derivatives off if they are not implemented, rather than set the derivative to zero (which results in the same answer but at the cost of doing work and then multiplying it by zero).


lipnikov (approved):
Fixed one XML issue. The rest looks good.

PR #759, opened 2023-07-17
Update amanzi-ci.yml to build images for arm64

rfiorella:
Updates the amanzi-ci.yml GitHub action to build images for both amd64 and arm64 using buildx. It is built on the TPLs image, which is currently only amd64 and will need to be updated accordingly.

Increments action versions and removes set-output for $GITHUB_ENV to remove deprecation warnings.

Currently only runs the tests on the amd64 image. I think the CI action could be modified to run on both architectures - let me know if this would be preferred and I'll update the PR. (Refs #762 and #760)


rfiorella:
Not sure this route will be viable without arm64 runners on GH actions. Closing this PR as a result.

PR #763, opened 2023-07-18
Fixed paralell copy of boundary_face component

lipnikov:
A boundary face_component created on a fly is not ghosted. However, field in the other state may be ghosted. This leads to different sizes of Epetra_MultiVectors in a state-based initialization.

### DIFF
--- a/src/data_structures/CompositeVector.cc
+++ b/src/data_structures/CompositeVector.cc
@@ -243,9 +243,11 @@ CompositeVector::operator=(const CompositeVector& other)
 
     if (Ghosted() && other.Ghosted()) {
       // If both are ghosted, copy the ghosted vector.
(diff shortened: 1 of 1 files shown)


ecoon (approved):

ecoon:
At some point we may need to take a closer look at boundary_faces across the code.  I'm not sure whether Rao and I were clear with whether boundary_faces needed to be ghosted or not, and they weren't always treated as a fully featured Entity_kind with both GHOSTED and OWNED maps, etc.  I think they are in the mesh refactor, but that may need to be checked systematically sometime.

lipnikov:
boundary_face component is connected with boundary conditions which are  based on ghosted objects. I think we reduce number of underlying assumptions and simplify logic of algorithms if we assume that this component looks like all over components.

PR #770, opened 2023-09-28
More work on units

lipnikov:
Changed definition of units in RecordSet to std::string. Added units output in statistics. Added optional additional compatibility hook for units and tested with in one evaluator used in Amanzi.

### DIFF
--- a/src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
+++ b/src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
@@ -152,7 +152,7 @@ EvaluatorMultiplicativeReciprocal::EvaluatePartialDerivative_(
 
 
 /* ******************************************************************
-* Compativility
+* Units are calculated if field has none. Otherwise, units are compared.
 ****************************************************************** */
 void
 EvaluatorMultiplicativeReciprocal::EnsureCompatibility_Units_(State& S)
--- a/src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
+++ b/src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
@@ -150,4 +150,31 @@ EvaluatorMultiplicativeReciprocal::EvaluatePartialDerivative_(
   }
 }
 
+
+/* ******************************************************************
+* Compativility
+****************************************************************** */
+void
+EvaluatorMultiplicativeReciprocal::EnsureCompatibility_Units_(State& S)
+{
+  std::string data("-");
+  Utils::Units system;
+  auto& r = S.GetRecordSetW(my_keys_[0].first);
+
+  for (auto it = list0_.begin(); it != list0_.end(); ++it) {
+    auto tmp = S.GetRecordSet(*it).units();
+    data = system.MultiplyUnits(data, tmp);
+  }
+  for (auto it = list1_.begin(); it != list1_.end(); ++it) {
+    auto tmp = S.GetRecordSet(*it).units();
+    data = system.DivideUnits(data, tmp);
+  }
+
+  auto tmp = r.units();
+  if (tmp != "")
+    AMANZI_ASSERT(system.CompareUnits(tmp, data));
+  else
+    r.set_units(data);
+}
+
 } // namespace Amanzi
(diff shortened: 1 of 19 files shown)


src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
```diff
@@ -150,4 +150,31 @@ EvaluatorMultiplicativeReciprocal::EvaluatePartialDerivative_(
...
 
+
+/* ******************************************************************
+* Compativility
+****************************************************************** */
+void
+EvaluatorMultiplicativeReciprocal::EnsureCompatibility_Units_(State& S)
+{
+  std::string data("-");
+  Utils::Units system;
```
ecoon, line 161: Is there ever a reason to use "non-default units"?  Would it be worth putting a Units system in State as a "global units system" that a user could change?

src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
```diff
@@ -150,4 +150,31 @@ EvaluatorMultiplicativeReciprocal::EvaluatePartialDerivative_(
   }
 }
 
+
+/* ******************************************************************
+* Compativility
```
ecoon, line 155: Compatibility (spelling typo)

src/state/evaluators/EvaluatorSecondaryMonotype.hh
```diff
@@ -135,6 +135,9 @@ class EvaluatorSecondaryMonotype : public EvaluatorSecondary {
                                            const std::vector<AmanziMesh::Entity_kind>& locations,
                                            const std::vector<int>& num_dofs)
   {}
+
+  // optional, anything for units
+  virtual void EnsureCompatibility_Units_(State& S){};
```
ecoon, line 140: Has this been run through clang-format?  I think clang-format puts a space here.

src/utils/Units.cc
```diff
@@ -275,14 +277,17 @@ Units::ConvertUnitD(double val,
...
 
 /* ******************************************************************
-* Convert unit string
+* Convert unit string 
 ****************************************************************** */
 std::string
 Units::ConvertUnitS(const std::string& in_unit, const UnitsSystem& system)
 {
+  if (in_unit == "") return "";
+  if (in_unit == "-") return "-";
```
ecoon, line 286: I like this.  Am I correct in assuming that "-" means "unitless" here, which is different from the empty string which means "not set"?

ecoon (changes requested):
Looks like it needs to get run through clang-format, but otherwise looks good to me.

I thought we used to use a library for units, but maybe that was boost and you've removed it previously?

Somewhere it would be nice to have a documentation of the syntax of units.  Is it "W/m^2"?  Is "W m^-2" ok?  Are prefixes e.g. "MJ" ok?  (Not required for this PR)

I've always wanted us to do a better job on this, so I'm glad you're working on it.  I'll add a ticket to make a push to add units to more of ATS at some point.

lipnikov:
formatting with clang-15 made no additional changes to the code.

Boost units were design to calculate with numbers having units. The underlying capabilities were not flexible enough to be useful for codes like Amanzi with the fixed unit system. Quite possible that since 2015 some libraries for units were developed. If so, we may consider them.

Syntax of units is too rigid now and does not look fancy. I'll add comments to the header and launch a discussion on GitHub. It is relatively easy to change it.  One design deficiency is that derived units like Pa are transformed to kg/m/s^2, so
units for complex fields in State (e.g. product of many fields) may be too long.

src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
lipnikov: By some user request, we allow to change from mol/m^3 to mol/L, other units are fixed. I do know about usefulness of a general unit system for calculating, currently it is used to transform units for I/O

src/state/evaluators/EvaluatorMultiplicativeReciprocal.cc
lipnikov: Fixed

src/state/evaluators/EvaluatorSecondaryMonotype.hh
lipnikov: Hmmm... I run again through clang 15 formatting. No change.

src/utils/Units.cc
lipnikov: You are correct.

Issue #773, opened 2023-10-19
shallow water tests are too long


ecoon:
Please take a look at the shallow water tests -- do we really need to run them for this long?  Maybe restart files could be used if you need to get key periods in time, or just run a short time window, or a smaller mesh?  These tests take a very large fraction of our CI runtime:

```
        Start 148: shallow_water_1D
148/221 Test #148: shallow_water_1D ...................................................................................   Passed  198.34 sec
        Start 149: shallow_water_2D_smooth
149/221 Test #149: shallow_water_2D_smooth ............................................................................   Passed   20.99 sec
        Start 150: shallow_water_bathymetry
150/221 Test #150: shallow_water_bathymetry ...........................................................................   Passed    0.42 sec
        Start 151: shallow_water_lake_at_rest
151/221 Test #151: shallow_water_lake_at_rest .........................................................................   Passed  350.36 sec
        Start 152: shallow_water_Thacker
152/221 Test #152: shallow_water_Thacker ..............................................................................   Passed   96.72 sec
        Start 153: shallow_water_dry_bed
153/221 Test #153: shallow_water_dry_bed ..............................................................................   Passed  1011.80 sec
```

gcapodag:
If I may chime in on this discussion, while working on the pipe model, I had to make a change to the way the boundary conditions are handled to ensure mass conservation. With that change, the lake at rest is now failing during continuous integration. @nvohra0016 we can talk about it at our meeting Thursday, but I thought mentioning it here was relevant to thread.

lipnikov:
The test are fast:  on OSX, I am getting this:

```
1/6 Test #1: shallow_water_1D .................   Passed    1.93 sec
2/6 Test #2: shallow_water_2D_smooth ..........   Passed    0.79 sec
3/6 Test #3: shallow_water_bathymetry .........   Passed    0.55 sec
4/6 Test #4: shallow_water_lake_at_rest .......   Passed    1.48 sec
5/6 Test #5: shallow_water_Thacker ............   Passed    6.61 sec
6/6 Test #6: shallow_water_dry_bed ............   Passed    2.91 sec
```

nvohra0016:
> If I may chime in on this discussion, while working on the pipe model, I had to make a change to the way the boundary conditions are handled to ensure mass conservation. With that change, the lake at rest is now failing during continuous integration. @nvohra0016 we can talk about it at our meeting Thursday, but I thought mentioning it here was relevant to thread.

[...]

PR #781, opened 2023-11-02
Update Dockerfile-TPLs to manually compile mpich

rfiorella:
Updates dockerfile for TPLs to automatically build mpich from source with custom options that dramatically speeds running of tests on github actions when test oversubscribes runners. Refs #773.

Benefits to CI require that a new TPL image with newly compiled mpich is pushed to dockerhub and picked up by future CI runs.


rfiorella:
New TPL image was generated last week - is this ready to merge or do we need more updates to the TPL dockerfile?

jd-moulton (approved):
The use of our own mpich seems like the best solution.  This seems to wrap things up and positions us for running with Shifter on NERSC.

PR #784, opened 2023-11-14
typo fix in nanoflann patch

rfiorella:


rfiorella:
No problem! Had some issues building the TPL images this morning after this patch was put on master while trying to debug mpich issues in one of the other open PRs.

PR #786, opened 2023-11-18
Full name of installed library

lipnikov:
Full name of the installed Hypre library.

### DIFF
--- a/config/SuperBuild/include/Build_HYPRE.cmake
+++ b/config/SuperBuild/include/Build_HYPRE.cmake
@@ -82,7 +82,8 @@ else()
 endif()
 
 # --- Set the name of the patch
-set(HYPRE_patch_file hypre-superlu.patch)
+set(HYPRE_patch_file hypre-superlu.patch
+                     hypre-configure.patch)
 # --- Configure the bash patch script
 set(HYPRE_sh_patch ${HYPRE_prefix_dir}/hypre-patch-step.sh)
 configure_file(${SuperBuild_TEMPLATE_FILES_DIR}/hypre-patch-step.sh.in
--- a/config/SuperBuild/templates/hypre-configure.patch
+++ b/config/SuperBuild/templates/hypre-configure.patch
@@ -1,0 +1,12 @@
+diff --color -ruNbB hypre-2.29.0/src/configure hypre-2.29.0-source/src/configure
+--- hypre-2.29.0/src/configure	2023-11-17 22:42:33
++++ hypre-2.29.0-source/src/configure	2023-11-17 22:36:17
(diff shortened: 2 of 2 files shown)


ecoon:
Is there a reason this is still here?  This would be appreciated so that we no longer have to set LD_LIBRARY_PATH all the time...

jd-moulton:
Failed to reproduce it, so was trying to understand why @rpath was potentially behaving differently on different systems.  But if it's leading to some people having to set the LD_LIBRARY_PATH, I'll accept it.  I just need resolve the conflict.

jd-moulton:
Tested this on my Mac (M2) running Ventura with Xcode 14.3 using Macports gcc 12.3 and locally built openmpi 5.0.1.   Although it isn't necessary on my system, it does work either way.  Here are a few notes/comments on this approach though, as it does seem to be evolving toward a standard use of @rpath in the future:

- CMake did update default behavior in version 3.0 and beyond [MACOSX_RPATH](https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_RPATH.html#prop_tgt:MACOSX_RPATH).  It favors the use of @rpath
- The use of @rpath in the HYPRE CMake seems consistent with the documentation on the CMAKE_MACOSX_RPATH variable and default setting.
- Using otool you can see if the libraries are being found and what their path is.  
- Current behavior is,
  ``` 
  moulton@pn2401338 bin % otool -L amanzi | grep HYPRE
	@rpath/libHYPRE-2.29.0.dylib (compatibility version 0.0.0, current version 0.0.0)
  ```
and  the @rpath can be found by using otool -l (pipe through less and search on RPATH), and I find in the list 
  ```
          cmd LC_RPATH
      cmdsize 104
         path /amanzi/tpls/install/OSX-13.6.7-arm64/openmpi-5.0.1-gcc-12.3.0/Release-0.98.9-shared/lib (offset 12)
  ```
- New behavior is,

  ``` 
  moulton@pn2401338 standalone_simulation_coordinator % otool -L amanzi | grep HYPRE
	/amanzi/tpls/install/OSX-13.6.7-arm64/openmpi-5.0.1-gcc-12.3.0/Release-0.98.9-shared/lib/libHYPRE-2.29.0.dylib (compatibility version 0.0.0, current version 0.0.0)
  ```
So you can see the explicit path.   This is more consistent with the other TPLs, but it may be that everything will get switched back over to @rpath in the future (if we can make sure it works for everyone).

jd-moulton (approved):
Changes are reasonable, and are consistent with the other TPLs.  But they shouldn't be necessary (see discussion in comments). But since for some people (likely meaning some XCode environments) it helps, we should go ahead with the patch.

PR #791, opened 2023-12-01
Changes UniqueMeshFunction to instantiate only on OWNED entities

ecoon:
This is required because UniqueMeshFunction is used on source terms, and those source terms are only well posed on OWNED entities when they are derived from subgrid meshes that are only instantiated on OWNED entities.

This may break some things -- run tests and confirm please CI.  If it doesn't break things, check with @lipnikov as to whether he is ok with this change.

### DIFF
--- a/src/mesh_functions/BoundaryFunction.hh
+++ b/src/mesh_functions/BoundaryFunction.hh
@@ -36,7 +36,8 @@ namespace Functions {
 class BoundaryFunction : public UniqueMeshFunction {
  public:
   BoundaryFunction(const Teuchos::RCP<const AmanziMesh::Mesh>& mesh)
-    : UniqueMeshFunction(mesh), finalized_(false){};
+    : UniqueMeshFunction(mesh, AmanziMesh::Parallel_kind::ALL),
+      finalized_(false) {};
 
   void Define(const std::vector<std::string>& regions, const Teuchos::RCP<const MultiFunction>& f);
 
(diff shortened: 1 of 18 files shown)


ecoon:
> @ecoon If subgrid mesh has no ghosts then ALL must be the same as OWNED.

Sorry, I wasn't clear.  The mesh on which the DomainFunction using this UniqueMeshFunction is a normal, parallel mesh (surface mesh).  On each owned cell, that mesh creates a LogicalMesh that is the subgrid mesh.  The DomainFunction integrates the effect of the sub grid mesh into a vector on the surface mesh.  As it is defined now, the DomainFunction expects to work with ALL cells, computing a source term on ALL cells.  I'm not sure why a source term would ever be computed on ALL cells -- we only use it on OWNED cells.  But the sub grid meshes which are integrated across to form the source only exist on each OWNED cell.

lipnikov:
I think UniqueMeshFunction is used for boundary regions too. The ghost entries are used in BCs to reduce the number of parallel communications. [Note that an owned cell may have ghost nodes.] If so, this change  breaks everything related to 
shared boundary objects (nodes, edges).

ecoon:
Based on the tests, it looks like only shallow water uses Ghosted entities for BCs?

I guess it should be straightforward to require a Parallel_kind when constructing a UniqueMeshFunction -- BCs could be ALL (or at least shallow water BCs could be all) and source terms could be OWNED.

lipnikov:
@ecoon Since "domain" could mean a boundary domain too, I think that parallel_kind should be specified for each PK_DomainFunctionXXX constructor.

ecoon:
@lipnikov  I think this is ready to go -- I've added a "ghosted = false" option to the PK_DomainFunctionFactory constructor, then updated each PK to use "ghosted = true" for the boundary conditions (and false for source terms).  This gets the shallow_water and richards_seepage tests to pass.

What I did not try to do is figure out which of the PKs/discretizations actually need their boundary conditions to be computed on ghost cells.  As  a result, a few PKs may being doing extra work (but no more than they did already on master).  Nor did I check to see if any source terms need ghosted entries.

All tests pass, let me know if you would like to see any other changes.

lipnikov:
Any scheme using nodal DoFs may need ghosted BC values. e.g. schemes in mechanics PKs.

lipnikov (approved):

Issue #792, opened 2023-12-01
TPL Updates for version 0.98.8


jd-moulton:
Our TPL updates (#742) slipped behind the update of Trilinos and Hypre (which became TPLs 0.98.7).  We need to recapture the updates of several libraries that were integrated in the old TPLs 0.98.7 branch.   This new ticket (and branch cut from current master) will capture that work.  

We need to aggregate updates for TPLs coming in from PRs to avoid users having to rebuild the tpls too often. We may integrate other minor updates to some libraries as well. The current PRs are:

- [x] - update ExprTk package (https://github.com/amanzi/amanzi/pull/734)
- [x] - patch netcdf-fortran cmake (https://github.com/amanzi/amanzi/pull/735)
- [x] - fix confusing naming (https://github.com/amanzi/amanzi/pull/736)

In addition, to prepare for the work on gas diffusion we need to update geochemistry related libraries

- [x]  alquimia v1.1.0
- [x]  PetSc v3.20.0
- [x]  Sowing (1.1.26-p8) that goes with PetSc
- [x]  PFLOTRAN v5.0.0

And we need to restore the ability to build either CrunchFlow or PFLOTRAN, but not require both.

Other updates are possible but may wait (e.g., boost).

jd-moulton:
**Note:** these updates will fail CI until I reach a point where it is worth updating the Docker TPLs image to match the new version.

ecoon:
This fails for me:  

CMake Error at tools/cmake/AmanziVersion.cmake:123 (string):
  string sub-command REGEX, mode MATCH needs at least 5 arguments total to
  command.
Call Stack (most recent call first):
  CMakeLists.txt:78 (include)

line 123 is:

`  string(REGEX MATCH "\\[(.*)\\]" _ ${AMANZI_GIT_PARENT_BRANCH_MSG})`

I'm not sure what you're trying to do here

ecoon:
For now I just commented out this line and the following line, and moved on....

ecoon:
Also this does not work for me:

AmanziTPL.cmake:26

```string(REGEX REPLACE ".*AMANZI_TPLS_VERSION_PATCH ([0-9]+)([a-z]+).*" "\\1\\2" AMANZI_TPLS_VERSION_PATCH_SRC "${line}")```

When the patch has no letters at the end, it does not match and I get:

```
-- >>>>>>>> AmanziTPL.cmake
[...]

PR #793, opened 2023-12-01
added natural_id to alquimia routine argument list

ghammond86:
This PR adds "natural_id" as an argument for calls to Alquimia, making Amanzi compatible with the Nov 23 xSDK versions of PFLOTRAN and Alquimia. natural_id was added for debugging purposes.

### DIFF
--- a/src/common/alquimia/ChemistryEngine.cc
+++ b/src/common/alquimia/ChemistryEngine.cc
@@ -558,7 +558,7 @@ ChemistryEngine::Advance(const double delta_time,
                          AlquimiaState& chem_state,
                          AlquimiaAuxiliaryData& aux_data,
                          AlquimiaAuxiliaryOutputData& aux_output,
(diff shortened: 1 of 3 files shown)


jd-moulton:
This was really helpful in patching the interface for the new TPLs branch.   The only issue was mesh_->cell_map, as the current implementation of the mesh cache doesn't have cell_map, instead it looks like the interface uses mesh_->getMap, in which you specify the Entity_kind.  See update under #792.

ecoon:
Is this blocked by TPLs needing to be updated?

jd-moulton:
I think this is best tied to the update of the TPLs, and it is fixed there.   So this PR may end up just going away.  The delay in the TPL update was waiting on update/review of accepted results for a few geochemistry tests.  Sergi just completed those so my plan is to push the udpates to the TPLs in the next few days.

PR #807, opened 2024-02-07
Naren/giacomo urban storm drains

nvohra0016:
Added "outward discharge" BC implementation which allows for a discharge value to be specified without apriori knowledge of the boundary outward normal direction.


gcapodag:
Looks good, thanks @nvohra0016

gcapodag (approved):

Issue #811, opened 2024-02-16
Fix poor naming conventions in PK_DomainFunction


ecoon:
In most of Amanzi and ATS, a **name** is a user-provided, arbitrary string identifying an **instance**, while a **type** or **kind** or **method** is a non-arbitrary, "magic keyword" identifying a **class**.

For instance, `PK::name()` returns the arbitrary, user-provided title of the parameter list used to construct the PK, while the "PK type" is not arbitrary, and must be in the dictionary of valid types.

`PK_DomainFunction` and friends use `PK_DomainFunction::name()` to identify the class with hard-coded strings, and don't ever save the actual name, which is their parameter list's name.

Proposal:

1.  Change `PK_DomainFunction::name()` --> `PK_DomainFunction::getType()` or `getMethod()`. Note this actually appears in a few places -- since `TransportDomainFunction` does not inherit from `PK_DomainFunction`.
2.  Add `PK_DomainFunction::getName()` and make it save and return `plist.name()`

So, for instance, for the following transport source term:

```
      <ParameterList name="source terms">
        <ParameterList name="component mass source">
          <ParameterList name="Tc-99 injection">
            <Parameter name="component names" type="Array(string)" value="{Tc-99}" />
            <Parameter name="regions" type="Array(string)" value="{rainfall patch}" />
            <Parameter name="spatial distribution method" type="string" value="none" />
            <ParameterList name="source function">
              <ParameterList name="function-constant">
                <Parameter name="value" type="double" value=" 1" />
              </ParameterList>
            </ParameterList>
          </ParameterList>
        </ParameterList>
      </ParameterList> 
```

Would result in:

```
PK_DomainFunctionSimple<TransportDomainFunction>::getType() --> "simple"
PK_DomainFunctionSimple<TransportDomainFunction>::getName() --> "Tc-99 injection"
```

I recognize this may require a significant number of changes in Amanzi physics, so I would like to get @lipnikov  buy-in before proceeding....

ecoon:
Alternate choices for `getType()` are `getMethod()` and `getSpatialDistributionMethod()` (since that is the parameter's name).  I don't much care which, though I would argue that spatial distribution method doesn't capture the full extent of the variations of what is implemented there.

lipnikov:
@ecoon  I'll think about this. Let us first resolve a simpler question: get_name() or getName(). One of your previous tickets #486 argue for enforcing standard for access/setup functions like get_name() and set_name(). I think we have not established a standard...

[...]

PR #813, opened 2024-02-23
Changed naming conventions in PK domain functions

lipnikov:
Renamed name() to getName() and added getType() which returns enum value.


ecoon:
refs #811

PR #815, opened 2024-03-04
cleans up a few names in writing checkpoint/vis files; stops use of pk name in tag

ecoon:
Removing spaces from some names for checkpoint/vis, removes use of PK name as a tag, which is inconsistent with how tags are used everywhere else.

### DIFF
--- a/src/time_integration/SolutionHistory.hh
+++ b/src/time_integration/SolutionHistory.hh
@@ -170,7 +170,7 @@ SolutionHistory<Vector>::Initialize_(int mvec, const Vector& initvec)
 
     // require time deltas
     for (int j = 0; j < mvec; j++) {
-      S_->Require<Vector>(initvec.Map(), name_, Tag(std::to_string(j)), name_);
+      S_->Require<Vector>(initvec.Map(), "solution_history_" + name_, Tag(std::to_string(j)), name_);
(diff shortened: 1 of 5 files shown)


src/time_integration/SolutionHistory.hh
```diff
@@ -170,7 +170,7 @@ SolutionHistory<Vector>::Initialize_(int mvec, const Vector& initvec)
 
     // require time deltas
     for (int j = 0; j < mvec; j++) {
-      S_->Require<Vector>(initvec.Map(), name_, Tag(std::to_string(j)), name_);
+      S_->Require<Vector>(initvec.Map(), "solution_history_" + name_, Tag(std::to_string(j)), name_);
```
lipnikov, line 173: Name has been changed in one place but PR passed. What does it mean? A quick comment would help...

lipnikov: Solution history is certainly used to extrapolate solution to get a better initial guess for a NKA. Probably this state change is needed for the perfect restart. Inside a single time period, the extrapolation algorithm uses vectors d_[I]. I think Amanzi does not have tests for a perfect restart, unless you added some recently. Hence PR passes.

lipnikov (approved):
Approval is contingent to passing all tests.

PR #823, opened 2024-04-03
Add pipe flow PK

gcapodag:
This PR adds a pipe flow PK, which solves the 1D Saint Venant equations (in a 2D domain). The pipe flow can be coupled during the time integration step with a junction model that solves the 2D SW equations, with an additional friction term compared to the current SW PK. Pipe flow + junctions form the pipe network, which can be coupled with an overland flow model, where water can move from the surface down into the network and back up, in case overflow occurs. The pipe network model has been tested and is mass conserving, and preserves steady state initial conditions, similarly to the SW model. Tests currently exist to provide continuous integration and will soon be included in an ats-regression-tests suite PR.
The pipe flow model in standalone mode has been tested against experimental data with the hydraulic routing test from [Liu and Chen](https://ascelibrary.org/doi/full/10.1061/(ASCE)HY.1943-7900.0002011), see pictures below, showing results that are consistent with those in the aforementioned paper. Please also see the companion PR in ATS for the pipe drain evaluator regulating the water exchange between pipe network and overland flow: https://github.com/amanzi/ats/pull/252
hyd1
hyd2


lipnikov:
I think it would be good to rebase this PR and squash the history, since many commits are failing. Thoughts?  Also, run it through Amanzi formatting script in tools/formatting to help the review process.

gcapodag:
Reopened here: https://github.com/amanzi/amanzi/pull/824

PR #825, opened 2024-04-03
work to optimize use of View in mesh construction/caching

ecoon:
- removes many Kokkos::resize() in favor of Kokkos::realloc(), which is significantly faster if you don't need to keep the old data
- avoids deep_copyies on host when a loop would suffice
- avoids deep_copies when subviews would suffice
- tries to take advantage of move semantics in View::operator=.  Not sure if this is significant, but since a move operator= is defined, we should use it where possible.

### DIFF
--- a/src/mesh/mesh_extracted/MeshExtractedManifold.cc
+++ b/src/mesh/mesh_extracted/MeshExtractedManifold.cc
@@ -324,8 +324,8 @@ MeshExtractedManifold::InitParentMaps(const std::string& setname)
     { Entity_kind::FACE, Entity_kind::EDGE, Entity_kind::NODE });
 
   for (int i = 0; i < 3; ++i) {
-    auto kind_d = kinds_extracted[i];
-    auto kind_p = kinds_parent[i];
+    Entity_kind kind_d = kinds_extracted[i];
+    Entity_kind kind_p = kinds_parent[i];
 
     // build edge set from Exodus labeled face set
     Entity_ID_View setents;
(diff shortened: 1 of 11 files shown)


ecoon:
Note, there is still room for improvement in Mesh construction -- most of the runtime is now spent in allocating small views in e.g. Mesh_MSTK::getFaceCells() and comparable.  In the old code, the API was:

`void getFaceCells(Entity_ID f, std::vector<Entity_ID>& cells);`

The point of this was to re-use cells in a loop over faces to cache the face-cell connectivity.

This breaks data parallelism, however -- you can't reuse the data in a threaded environment, so we changed this to:

`void getFaceCells(Entity_ID f, Kokkos::View<const Entity_ID*, HOST> cells);`

Because you can't assign to cells, this forced the developer to follow a paradigm of:

- alloc a view of non-const Entity_ID, `Kokkos::View<Entity_ID*, HOST> lcells`
- populate that view
- assign: `cells = lcells` which is a shallow assignment, changes the view but not the entries of the view.

This does not (cannot) share the memory, and as a result, we do a ton of small allocs.


I think we could continue to improve runtimes by changing the API to 

`void MeshFramework::getCellFaces(Entity_ID f, Kokkos::View<Entity_ID*, HOST>& cells);`

This breaks data parallelism, but only on HOST and only on MeshFramework, but would allow us to reuse the cells view in the loop over faces.

Passing in a `View<Entity_ID*>`, however, means that either:

1. The API differs between:
   `MeshFramework::getFaceCells(Entity_ID, View<Entity_ID*>&)` and 
   `MeshCache::getFaceCells(Entity_ID, View<const Entity_ID*>&)`

OR

2. We keep MeshCache and MeshFramework the same, but make `MeshCache::getFaceCells(Entity_ID, View<Entity_ID*>)` only valid on HOST.  This would force DEVICE functions to use the `View<const Entity_ID*, DEVICE> = MeshCache<DEVICE>::getFaceCells(Entity_ID f);` version.


1 would be obnoxious because things like MeshAudit and some MeshAlgorithms are templated on Mesh_type, and are expected to work for both MeshCache and MeshFramework, and making MeshFramework and MeshCache APIs different would break those.

2 would be obnoxious because it means we have more differing API between MeshCache<HOST> and MeshCache<DEVICE>.  But maybe this is not so bad if we are already splitting these implementations to avoid false-positive warnings.  And it may be obnoxious because we may have old code that uses the old API, and we'd have to refactor that to use the "return view" API instead of the "pass view by reference" API.  But this is probably the right thing to do anyway (as DEVICE code should use the CACHED algorithm anyway and that wants to return the View.



The downside of this is that

src/mesh/mesh_extracted/MeshExtractedManifold.cc
```diff
@@ -333,16 +333,16 @@ MeshExtractedManifold::InitParentMaps(const std::string& setname)
     if (setents.size() == 0) {
       auto [lsetents, lvfs] =
         parent_mesh_->getSetEntitiesAndVolumeFractions(setname, kind_p, Parallel_kind::ALL);
-      Kokkos::resize(setents, lsetents.size());
-      Kokkos::deep_copy(setents, lsetents);
+      Kokkos::realloc(setents, lsetents.size());
+      for (size_t i = 0; i != lsetents.size(); ++i) setents(i) = lsetents(i);
```
ecoon, line 337: something wrong here?  At the least can't use i here, it is used above in 326.  Also, can't we just use lsetents directly and no setents view?  What does the loop over i do here?

PR #828, opened 2024-04-05
fixes bug in vandelay vector being hard-coded as only-owned

ecoon:
also adds better testing of scattering boundary_faces


ecoon:
@lipnikov  can you test your case for ghosted boundary faces based on this branch?  The code was all there for it in Mesh, but the vandelay map was hard-coded to owned in CompositeVector.

PR #830, opened 2024-04-08
Change to use Hypre directly, remove Ifpack2 interface. 

JulienLoiseau:
- Remove Ifpack2 from PreconditionerHypre
- Directly target features from Hypre
- Avoid problems from Ifpack, since the function copying matrices (to GPU) has issues in Ifpack2 and needs to be patched
- Add support for MRG as preconditioner, which is not supported through Ifpack2. 

This interface is using Epetra, since our master branch is still based on it. This is an adaptation of the Tpetra interface for Hypre used in the [tpetra](https://github.com/amanzi/amanzi/tree/tpetra) branch. 

Todo: remove patch for Trilinos Ifpack2 and Ifpack.


jd-moulton (approved):
Great, looks like a natural extension of existing code to include MGR.  Konstantin caught the one typo and so everything is passing.

PR #833, opened 2024-04-23
changes clang-format to AllowShortBlocksOnASignleLine=false

ecoon:
I'm not sure how this ended up as our default formatting, since I think we both complained about it.  It makes debugging almost impossible because I can't set a breakpoint inside a for loop/if statement/etc.

Are you ok with this change?  It also updates the .clang-format file to make sure that it stays this way...

### DIFF
--- a/src/.clang-format
+++ b/src/.clang-format
@@ -9,7 +9,7 @@ AlignEscapedNewlines: Right
 AlignOperands:   true
 AlignTrailingComments: true
 AllowAllParametersOfDeclarationOnNextLine: false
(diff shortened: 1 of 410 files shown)


lipnikov:
Changing if (condition) {  found = false; } looks good to me due to brackets {}. 
What about if (condition) found = false; I think this should be in one line, which is also fine with me.

ecoon:
Yes, I agree with both.  I'll check, then rerun clang-format (rather than deal with merges) and commit.

lipnikov (approved):

PR #837, opened 2024-05-13
fixes bug in how we download Trilinos -- use release not git

ecoon:
We used to need a specific hash of Trilinos -- currently we are using Trilinos 15.1.0.

When we stopped using the old hash, I must have missed that changing the version wasn't enough -- because we provided a GIT_REPOSITORY but not a GIT_HASH, we were likely getting the current master of Trilinos.

This changes Build_Trilinos to use the release version, not current master.


ecoon:
Note I have no clue if this breaks CI... maybe not because CI was probably built when current master Trilinos was basically 15.1

jd-moulton (approved):
I agree that using the git repo reliably would have required the hash of the release.   Fixing this could have simply added the hash, however, I also agree that using official releases whenever possible is a better practice.  So I think this is the best solution for now.

PR #848, opened 2024-07-20
add elm_ats_api build option to boostrap

rfiorella:
Adds the "--elm_ats_api" flag to bootstrap to optionally turn on building the ATS interface to ELM. Default false.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -140,10 +140,10 @@ unstructured=$TRUE
 geochemistry=$TRUE
 amanzi_physics=${TRUE}
 ats_physics=${FALSE}
-clm=${FALSE}
 
 # -- ats components
 elm_ats_api=${FALSE}
+clm=${FALSE}
 
 # -- mesh frameworks
 #    stk framewotk was deprecated and removed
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -577,6 +584,9 @@ Amanzi Components:
     clm            = '"${clm}"'
```
ecoon, line 584: can you put clm under ATS Components as well?

bootstrap.sh
```diff
@@ -718,6 +728,10 @@ List of INPUT parameters
                  ats_dev=${TRUE}
                  ;;
 
+      --elm_ats_api)
```
ecoon, line 731: So this should probably go under the logic for "--enable", so the flag should be:

--enable-elm_ats_api

If you go through and follow the same pattern as clm it should be right.

bootstrap.sh
```diff
@@ -2016,6 +2030,7 @@ cmd_configure="${cmake_binary} \
     -DENABLE_OpenMP:BOOL=${openmp} \
     -DENABLE_AmanziPhysicsModule:BOOL=${amanzi_physics} \
     -DENABLE_ATSPhysicsModule:BOOL=${ats_physics} \
+    -DBUILD_ATS_ELM_API:BOOL=${elm_ats_api} \
```
ecoon, line 2033: Same as above, this should probably be:

-DENABLE_ATS_ELM_API

for consistency's sake.

ecoon (changes requested):

ecoon (approved):

Issue #801, opened 2024-01-17
return error codes and deal with missing variable names


ecoon:
HDF5_MPI currently has no error handling, really because ASCEMIO has no error handling.  If a user causes the code to ask for an attribute or field that does not exist in an HDF5 file, HDF5 prints a bunch of stuff somewhere but there is no error code, no thrown error, etc.  So there is no way for Amanzi to detect that HDF5 was unable to find the thing, and therefore no way to throw a coherent error message.

This could be fixed by either fixing ASCEMIO or removing it entirely.

ecoon:
@lipnikov  is this related to #799 ?

lipnikov:
@ecoon It is tangent. In #799, the name of a field was mis-formed inside C++ code, but the initialize-from-file scenario does not require that a field be present. In other words, this is a partial initialization of state.

lipnikov:
@ecoon is this related to #76? If so, can we close the old ticket?

ecoon:
Ok, this really bit Bo and me recently -- in a big parallel intermediate scale model there was a bug in the attribute names, which were then being read from the incorrect checkpoint file, which resulted in... no errors because ascemio doesn't deal with errors.  This is really just plain stupid.

I have a fix for ascemio to at least error when reading attributes.  I may fix it for vectors as well.  How are we dealing with ascemio?  Is there a repo somewhere that I can push code to?  Or do I just add more patches to the TPLs?

ecoon:
> @ecoon is this related to #76? If so, can we close the old ticket?

yes, I closed that ticket as this one is "bigger" than that one and fixing this will fix that.

PR #858, opened 2024-08-16
Fixes a few long-standing issues with errors, including hdf5 errors

### LINKED ISSUE #743: “what():  Field  does not exist in the state”
A recent run gave the error quoted in the title.

This error does not help -- why is there no name written?  Or is there an empty string in there?

ecoon:
* fixes poorly communicated error messages in state (closes #743)
* renames `src/output/test/test_*.cc` --> `output_*.cc` to follow conventions in test naming for the rest of Amanzi
* introduces, along with a not-yet-provided patch of ASCEMIO, error checking in HDF5_MPI -- this means that incorrect files accessed when reading both attributes and vectors in checkpoint/restart, variables that read attributes or fields from files as a matter of initialization or "from file" evaluators will now throw errors rather than just printing hdf5 warnings to the screen, refs #801  (will close it once patch is in)
* adds tests to confirm that missing attributes, mistyped attributes (e.g. requesting an int as a double), missing fields, and incorrect-dimension fields all throw errors

NOTE: this is WIP because it requires changes to ASCEMIO, which need some input from @jd-moulton on whether to patch in SuperBuild or to make a "new release" of ASCEMIO or what.  Even without that patch, it introduces some error checking.


ecoon:
Ok, ASCEMIO patch is now updated with more stuff to deal with errors in attributes.

ecoon:
Ok, the remaining test is failing as expected due to error messages not being thrown by ASCEMIO.

ecoon:
@jd-moulton  sorry, I needed to tweak the patch to fix the tests, so TPLs need to be rebuilt.

rfiorella:
Ref #874, #875 - probably need ba9e774 for TPL rebuild

ecoon:
What do we need to push this out the door?  Any guesses why flow_darcy_well is failing?

lipnikov:
@ecoon Error message is currently printed right after the failed test: something has changed regarding treatment of input parameters in .h5 files.

Issue #646, opened 2022-02-14
Problems in current use of Time Tags and PK API


ecoon:
There are a few problems with the PK API and our current design of tags that make it hard to do a few things.

1. We have no clean way for a PK which needs an arbitrary field at both the old and new times to stash copies of that field.  For instance, transport needs liquid saturation, water content, and maybe a few other things at the old and new times.  It may even subcycle; therefore it needs to interpolate between old and new times.  ~Currently the flow PK is responsible for stashing the copy of liquid saturation at the old time, and copying it into the old time tag upon CommitStep().~  edit: this is incorrect, and is part of my confusion... if it is transport's job, not flow's job, to manage the copy, this isn't a problem.

But what if saturation liquid is a function rather than provided by a flow solve?  Do we save the copy?  Do we simply make two independent variable evaluators?

This is a problem in deformation as well, which uses saturation_{liquid, gas, ice} and porosity at the old time.  Who is responsible for copying these over?

2. In subcyling, we need four conceptual copies -- my_subcycle_current, my_subcycle_next, CURRENT, and NEXT -- and three actual copies (as NEXT and my_subcycle_next can share memory).  Who knows all the primary variables and requires the copies for CURRENT/NEXT?  Right now, the PK that is subcycled will require them for my_subcycle_next and my_subcycle_current, but there is no way of writing a generic subcycling MPC that requires the right primary variables at the outer interval.

lipnikov:
Bullet 1 is not clear. Each PK knows the start and end tags. It is also knows tags for all external fields it uses, e.g. from an MPC or input spec.  It is now up to this PK to use algorithms that preserve the State. Quite clear objective in my opinion.  If we know all external tags, a new evaluator could  created at a new tag and passed down to sub-PKs. What is wrong?

ecoon:
To be clear, I'm not sure 1 needs new Amanzi development, but it does need some coordination about how PKs are supposed to deal with time tags, so I wanted to discuss it.

Consider the simplest case of this -- transport needs liquid saturation.

I agree that transport knows about CURRENT, NEXT, and it may or may not have been given an addition set of tags, SUBCYCLE_CURRENT and SUBCYCLE_NEXT. 

It must be able to get saturation at its current and next times, whether that is the global tags CURRENT/NEXT or the local tags SUBCYCLE_CURRENT/SUBCYCLE_NEXT.  There are several cases, from most to least common:

[...]

Issue #673, opened 2022-06-22
figure out the right way to set the water tags for reactive transport


ecoon:
Currently Chemistry_PKs have hard-coded their use of Tags::NEXT for saturation, porosity, and density (see Alquimia_PK::CopyToAlquimia() method).

Transport may use either Tags::NEXT, if there is no subcycling, or tag_subcycle_next_, which could be set either internal or external subcycling.

ecoon:
Likely fixed when #646  is fixed

PR #860, opened 2024-08-26
Rework of PKs to split out parseParameterList, make SetEvaluator() call EnsureEvaluators()

### LINKED ISSUE #654: broken concept: RequireDerivative
This ticket is a placeholder for future work.

Currently Evaluator::IsDifferentiableWRT() is not valid until the entire sub-dag rooted at Evaluator is completed, which is done in EnsureEvaluators().

However, the current ordering is:

```
PK::Setup()
  - calls S::RequireEvaluator()
  - calls S::RequireDerivative()
S::Setup()
  - calls EnsureEvaluators()
```

This is clearly flawed -- what happens when a PK doesn't know if a derivative is valid.  For instance, a PK may get an independent variable for a source Q, or it may get a secondary variable for Q that depends upon its primary variable p.  In the former case, there is no "dQ/dp", and `GetEvaluator(Q)::IsDifferentiableWRT(p)` would return false.   In the latter case, there is a "dQ/dp" and `GetEvaluator(Q)::IsDifferentiableWRT(p)` would return true.  The PK developer should be able to use this to determine whether to require the derivative or not, but cannot do so because EnsureEvaluators() has not been called.

Fix this after state!

ecoon:
This creates another level of setup, called parseParameterList(), that occurs after PK construction but before Setup. Most ATS PKs constructor contents are now moved into parseParameterList(), and all PKs are modified to have them set their primary variables in this call. This allows the key advancement that calling State::SetEvaluator(key, tag, eval) (the function that is called inside of RequireEvaluator(...) after construction) can now call eval->EnsureEvaluators(). This means that the DAG is now valid as soon as the evaluator is set. This allows some important changes to the use of the dag, e.g. calling Evaluator::IsDifferentiableWRT() immediately after calling RequireEvaluator().

This closes #654  and is an important step toward #646

### DIFF
--- a/src/mpc/CycleDriver.cc
+++ b/src/mpc/CycleDriver.cc
@@ -1055,8 +1055,9 @@ CycleDriver::ResetDriver(int time_pr_id)
   //if (observations_ != Teuchos::null) observations_->RegisterWithTimeStepManager(tsm_);
 
   // Setup
-  pk_->Setup();
   pk_->set_tags(Tags::CURRENT, Tags::NEXT);
+  pk_->parseParameterList();
+  pk_->Setup();
 
   S_->Require<double>("dt", Tags::NEXT, "dt");
   S_->Setup();
--- a/src/mpc/CycleDriver.cc
+++ b/src/mpc/CycleDriver.cc
@@ -228,6 +228,7 @@ CycleDriver::Setup()
     }
   }
 
+  pk_->parseParameterList();
   pk_->Setup();
   S_->Require<double>("dt", Tags::NEXT, "dt");
   S_->Setup();
--- a/src/pks/CMakeLists.txt
+++ b/src/pks/CMakeLists.txt
@@ -36,11 +36,6 @@ include_directories(${ASCEMIO_INCLUDE_DIR})
 include_directories(${Epetra_INCLUDE_DIRS})
 include_directories(${HDF5_INCLUDE_DIRS})
 
-if(ENABLE_ATSPhysicsModule)
-  add_definitions(-DATS_DAG_POLICY)
-endif(ENABLE_ATSPhysicsModule)
-
-
 #
 # Library: pks
 #
--- a/src/pks/CMakeLists.txt
+++ b/src/pks/CMakeLists.txt
@@ -36,6 +36,11 @@ include_directories(${ASCEMIO_INCLUDE_DIR})
 include_directories(${Epetra_INCLUDE_DIRS})
 include_directories(${HDF5_INCLUDE_DIRS})
 
+if(ENABLE_ATSPhysicsModule)
+  add_definitions(-DATS_DAG_POLICY)
+endif(ENABLE_ATSPhysicsModule)
+
+
 #
 # Library: pks
 #
--- a/src/pks/PK.hh
+++ b/src/pks/PK.hh
@@ -127,6 +127,12 @@ class PK {
   // Virtual destructor
   virtual ~PK() = default;
 
+  // call to allow a PK to modify its own list or lists of its children.
+  virtual void modifyParameterList() {}
+
+  // read said list
+  virtual void parseParameterList() {}
+
   // Setup
   virtual void Setup() = 0;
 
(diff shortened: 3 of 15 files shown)


ecoon:
@lipnikov  I'm working toward closing all of the following:

amanzi/amanzi#654
amanzi/ats#167
amanzi/amanzi#673
amanzi/amanzi#646
amanzi/ats#186

Effectively this is building toward using tags more consistently across PKs, whereby each PK gets its tags by a call to `PK::set_tags()`, allowing generic subcycling MPCs to create "sub-tags" that can be used by the PK to subcycle.  This removes a bunch of hard-coded tag names in ATS's transport, chemistry, and other places.

Examples of something like this works in ATS's intermediate scale runs (flow + energy only) which subcycle individual columns relative to each other.

One of the tricky things about this is making sure we have a way of moving quantities from e.g. saturation@subcycled_next_tag --> saturation@NEXT, the global NEXT tag.  These can be "aliases" in the sense that they can share memory.  Examples of needing this are things like an Observation needing a subcycled quantity -- observations are always done at NEXT, but the subcycled quantity may not exist at NEXT.  So we create an "alias" evaluator and vector that points from e.g. TCC@NEXT --> TCC@subcycled_next.

Also, generalizing the use of Tags for subcycling requires taking the time interpolation of flow fields out of Transport and into an evaluator.  So, we're introducing "time interpolation" evaluators that interpolate across tags (e.g. saturation@transport_tag_next interpolates saturation@NEXT and saturation@CURRENT).

The challenge then is knowing how to get things like this that are interpolated.  For instance, consider saturation, which transport interpolates from NEXT --> transport_subcycle_next.  Clearly, saturation@NEXT must be a real evaluator computed as a function of pressure@NEXT.  It can't be an alias pointing back to saturation@transport_subcycle_next.  But the code doesn't know the difference between the two use cases.

The key to fixing this is to make sure that the full DAG is being constructed as we go.  What I mean by that is that, if flow calls RequireEvaluator("water_content", NEXT), then this will immediately also RequireEvaluator("saturation", NEXT), and we'll get the right evaluators (no aliases).  To do this, we need a PK to set that "pressure" is a "primary variable" when it first reads the parameter list.

This PR takes the first step toward this.  It breaks Amanzi, so I've "ifdef'd" the key change -- if ATS_DAG_POLICY is defined true, then calling SetEvaluator() on any evaluator immediately calls EnsureEvaluators(), meaning that the sub-tree rooted at this evaluator is constructed and is now valid.  This also helps with things like amanzi/ats#167, so you can say, 

`RequireEvaluator(source_term); if GetEvaluator(source_term).IsDifferentiableWRT(primary_var), then RequireDerivative(source_term, primary_var);`

So PKs can know if the source term is a function of the primary variable immediately, and require derivatives for use in the Jacobian.

In ATS, this required a few things:

1. added a PK::parseParameterList() function which, along with parsing the local options (as used to be in the constructor), it now also sets evaluator lists for all primary variables.  In order to do this, we need to have set_tags() called before parseParameterList() to know what tags we're using our primary variable at.  So it must be split from the constructor.
3. Make sure that all PKs that set up the DAG manually (eg. by constructing a specific evaluator and then calling SetEvaluator() manually, not through RequireEvaluator()) should instead put the "evaluator type" and all needed input spec in the input file during parseParameterList() and call RequireEvaluator instead.  For example, in Flow_PK.cc:170, you manually create the MultiplicativeEvaluator for wc_key.  This would break if any other evaluator depended upon water content and called RequireEvaluator() before this PK's setup was done.  Maybe this wouldn't be a big problem (flow is usually done first) but I'm not sure.

I'd like to get rid of that ifdef, but I'm not sure how many Amanzi PKs this will break.  And it breaks nearly any test that doesn't use the CycleDriver, because none of those tests call parseParameterList().

So I'll try and fix the tests, but I'm not sure how many will still break because the PK needs to be updated.

ecoon:
Ok, looking more closely at Amanzi's Flow_PK and Richards_PK, it seems you do a lot of manually constructed evaluators.  I'm not sure it is possible to order them so that the DAG will always be valid.  I don't see another way around this other than ifdef then...

PR #861, opened 2024-08-28
Build h5py using TPL HDF5 installation

rfiorella:
Moves h5py installation to after the tpl build, and always builds from source, so that h5py can use the tpl version of hdf5 instead of getting an additional copy through pip.


jd-moulton (commented):
@rfiorella Can you remind me what --no-binary=h5py does?

jd-moulton:
Tested this PR on my x86 MacBook Pro and it works for me.  However, I noticed the --help message for the deploy_* scripts doesn't describe all the options.  In parcticular, it doesn't describe the option for a proxy (--use_proxy).  Of course this is actually hard coded to be the LANL proxy so I was also thinking that we could have an option --use_LANL_proxy that uses the preset values in the script, and another option, say --use_ENV_proxy (or something like that) to use the https_proxy and http_proxy environment variables to set the proxy.   This way someone out side of LANL that has a proxy has an easy way to set it.

rfiorella:
> @rfiorella Can you remind me what --no-binary=h5py does?

This forces pip to ignore an existing wheel and build from source

rfiorella:
> Tested this PR on my x86 MacBook Pro and it works for me. However, I noticed the --help message for the deploy_* scripts doesn't describe all the options. In parcticular, it doesn't describe the option for a proxy (--use_proxy). Of course this is actually hard coded to be the LANL proxy so I was also thinking that we could have an option --use_LANL_proxy that uses the preset values in the script, and another option, say --use_ENV_proxy (or something like that) to use the https_proxy and http_proxy environment variables to set the proxy. This way someone out side of LANL that has a proxy has an easy way to set it.

I'll update this. For the proxy, I think the simplest way to gain this flexibility would be to just pass the http_proxy env variable to the script.

rfiorella:
@jd-moulton this seemed fixed now when I tried locally

Issue #685, opened 2022-08-30
bug in TimeStepManager?


ecoon:
This line of code seems suspicious to me:  https://github.com/amanzi/amanzi/blob/master/src/utils/TimeStepManager.cc#L122

My understanding is that this branch of the code is intended to avoid small "next" timesteps.  So for instance, if the time til the next "event" is 1s, and the timestep is 0.9 seconds, we would rather take two 0.5-second steps than a 0.9 + a 0.1.

The return statement looks right to me, e.g. https://github.com/amanzi/amanzi/blob/master/src/utils/TimeStepManager.cc#L127

I'm confused by what is being reserved here, and why.  

In this step, `dt_stable_storage = 2*dT - 0.5*time_remaining` (line 122).   We know that `dT > 0.75*time_remaining` (the condition), meaning that `2*dT > 1.5*time_remaining` meaning that `2*dT - 0.5*time_remaining = dt_stable_storage > time_remaining`. 

Then, at the next timestep, dt_stable_storage will always be values that ought to be limited by the next event (as time_remaining will now be half the previous value).  dT is assigned to be dt_stable_storage, and we will always hit the dT > time_remaining condition branch (line 113).  Now dt_stable_storage is assigned to be dT (this doesn't change anything) and we continue with dt as the time remaining, as expected.

Finally, in the next timestep (the first step after the event), dT is assigned, once again, to dt_stable_storage.  Presumably this is not limited by an event, so we will take this step, with a value of `2*dT - 0.5*time_remaining.`

My concern here is whether this is valid for the PK?  In particular, one would expect, since `dT > 0.75*time_remaining`, clearly `2*dT - 0.5*time_remaining > dT`.  So we're asking the PK to take a step larger than the one it last said was valid. 

So there are two questions here really:

1. Is this code correct?  Am I missing something?  I definitely think we should not ask the PK to take a step more than epsilon bigger than what it has said is valid, and this is definitely bigger than that.
2. Is storing dt_stable_timestep code necessary?  I believe the PK should be responsible for storing its timestep, and in the case where an event limits the actual timestep to less than is valid for the PK, and the PK happily solves that timestep with no trouble (not surprisingly), the PK can keep asking for bigger timestep sizes.  This is currently implemented in ATS.
3. If it is necessary, shouldn't this just get set to dT at line 122?  This will follow the same path through the code, and in the first step "after" the event, we would use the saved value, dT.

[...]

PR #863, opened 2024-09-06
WIP: Fixes problems with TimeStepManager

### LINKED ISSUE #821: TimeStepManager and transport
TimeStepManager stores a "dt_stable_storage", which is intended for use in the following situation:

1. Flow admits large dt, let's say 10 days.
2. Events (e.g. checkpoint request) limits this to let's say 1 day.
3. In the next get_dt() call, because flow only successfully took a 1 day step, it now "votes" for 1 day.
4. Then, in the TimeStepManaager, if the last step was limited by an event, rather than listening to what flow actually voted for (1 day) we use dt_stable_storage = 10 days.

This is needed to recover the "good" step size because Flow PK isn't smart enough to realize that despite only taking a 1 day tilmestep, it could happily have taken a 10 day step.

This breaks in the case of transport.  Currently, I am seeing the following:

1. Transport's initial step has 0 flux, so CFL has no limitation, so the initial dt provided by transport is "initial time step [s]" provided by the user.  Let's say this is large, e.g. 10 days.
2. Events limit this to 1 day.
3. In the next get_dt() call, transport's CFL kicks in and votes for 1 hour.
4. TimeStepManager thinks it knows that transport can take a 10 day step, and overrides the 1 hour vote by transport in favor of 10 days.

In the test I'm debugging, this gets chained -- every single step is limited by the 1 day checkpoint event, so every single step transport votes for 1 hour due to CFL, which is then overwritten by dt_stable_storage to 10 days, which is then limited to 1 day, and every single step is 1 day (bigger than transport's CFL).

This is more problematic version of the (unfixed) problem reported here: https://github.com/amanzi/amanzi/issues/685#issuecomment-1289085020 

I would like to change the code to do the following:

1. remove dt_stable_storage from the TimeStepManager
2. Let each PK keep track of both the dt it requested and the dt it received from the driver -- if it was limited and it thinks it can keep the bigger dt, then it should do that itself.

This would fix both problems.

The end result is that

`TimeStepManager::TimeStep(T, dT, after_failure) {}`

should _never_ return a value bigger than dT.

ecoon:
closes #821 and #685

Note this definitely broke some ATS regression tests.

It may break some Amanzi regression tests as well, due to differing timestep history.


ecoon:
@lipnikov  Note that this may require some changes to Amanzi PKs.  See, for example, my changes to Amanzi's Richards_PK.cc for an example of what I mean.

Specifically, if a PK recommends a large step size (say 1 day), and an event limits that size to a small step size (say, 1 hour).  Then, the PK may succeed, and even grow the step size (to maybe 1.5 hours), but then future steps after the event are also limited to the reduce step size (1.5 hours) and may not grow back to the valid PK step size (1 day).  This is due to the elimination of dt_stable_size from TimeStepManager (for reasons in the bugs referenced above).  I have provided an example fix for Richards PK which should work, conceptually, in other PKs as well.  Effectively the PKs need to keep a record of the step size they recommended and, if the actual step size taken is smaller than that recommendation and all goes well with the solve, then one should not reduce the recommended step size for the next step.

lipnikov:
@ecoon Time step controllers were designed to help a PK in tracking time steps. I am trying to understand why we still need to do time step management inside a PK. The logic regarding time events is quite generic to be moved to a controller. Thoughts?

src/utils/Event.hh
```diff
@@ -0,0 +1,147 @@
...
+  {
+    if (period_ <= (Scalar)0) {
+      Errors::Message msg("EventSPS provided invalid period.");
+      Exceptions::amanzi_throw(msg);
+    }
+    if (stop_ > (Scalar)0 && stop_ <= start_) {
+      Errors::Message msg("EventSPS provided invalid start/stop.");
+      Exceptions::amanzi_throw(msg);
+    }
+    if (start_ < (Scalar)0) {
```
lipnikov, line 106: It would be good for a supporting capability to not assume anything about time flow in events or physics. I understand that it is probably overkill for geophysics applications.

src/utils/Event.hh
```diff
@@ -0,0 +1,147 @@
...
+// Either int or double
+//
+template<typename Scalar>
+struct Event_EPS {
+  static constexpr int value = 0;
+};
+
+template<>
+struct Event_EPS<double> {
+  static constexpr double value = 1.e-12;
```
lipnikov, line 30: Is this a physics-based constraint or a numerical round-off constraint? If the first one, we probably need a way to set it up. If the second one, why is it so large?

src/utils/TimeStepManager.cc
```diff
@@ -16,109 +16,108 @@
...
 }
 
-TimeStepManager::TimeStepManager(Teuchos::ParameterList& verb_list)
+TimeStepManager::TimeStepManager(Teuchos::ParameterList& plist)
+  : manual_override_(false)
 {
-  dt_stable_storage = -1.;
-  vo_ = Teuchos::rcp(new VerboseObject("TimeStepManager", verb_list));
+  // manual override
+  if (plist.isParameter("prescribed timesteps [s]")) {
```
lipnikov, line 39: SI is the default unit system in Amanzi. Hence, we should use unit only for exceptions.

src/utils/test/utils_time_step_manager.cc
```diff
@@ -16,7 +16,7 @@
 
 TEST(TIME_STEP_MANAGER)
```
lipnikov, line 17: Many problems we had before were related to small dt. Do we have tests for such events?

ecoon:
> @ecoon Time step controllers were designed to help a PK in tracking time steps. I am trying to understand why we still need to do time step management inside a PK. 

That's probably right, we could pretty straightforwardly move the "stable step" size into the TimestepController, and as long as all PKs use the TimeStepController class, it should be fine.  I'll think about this a bit.

> The logic regarding time events is quite generic to be moved to a controller. Thoughts?

Events are outside of PKs -- they only need to live in one place (the cycle driver).

But I agree that if an event limits the step size, probably the logic to ignore that reduction in step size can be moved into the TimeStepController rather than in the PK.

src/utils/test/utils_time_step_manager.cc
ecoon: I think we have a test that checks if a dt is 0.99 * the time remaining, that we instead take 1/2 the time remaining.

I'm not sure we have much else.  I'm happy to start adding more tests in utils_time_step_manager.cc if you have ideas of corner cases that have broken.  I could mock up a test that demonstrates a few common failures and add them as well.

src/utils/TimeStepManager.cc
ecoon: I'll add both -- I've taken to preferring to use the unit everywhere because it makes the input spec self-documenting.  It's easy enough to tell people "all things are SI unless otherwise specified" but its even easier to not tell them and have it in the name.

Eventually I think it would be good to stop using ParameterList entirely and expand upon Teuchos ParameterList to include units (e.g. `<Parameter name="..." unit="..." value="..."/>` ) in the native spec.

src/utils/Event.hh
ecoon: It's what was in the code previously.  I think it's a numerical round-off contraint.  It could probably be a few orders smaller, I haven't tried.

src/utils/Event.hh
ecoon: You're saying negative start times are ok with you?  I'm ok with that.

ecoon:
Note this also closes #865

PR #867, opened 2024-10-01
Modified IO for fields with heterogeneous DoFs

lipnikov:
This is a work-around for visualizing fields with multiple DoFs per node. Complete solution is not feasible inside our mesh framework. This workaround gives nice pictures of nodal fields instead of a total mess.


src/output/OutputXDMF.cc
```diff
@@ -78,13 +90,7 @@ OutputXDMF::WriteMultiVector(const Epetra_MultiVector& vec,
 {
   AMANZI_ASSERT(names.size() == vec.NumVectors());
   for (int i = 0; i != vec.NumVectors(); ++i) {
-    if (mesh_->isLogical()) {
-      io_->writeNodeDataReal(*vec(i), names[i]);
```
ecoon, line 82: Is there a reason this is removed?  In logical meshes, there are no nodes, so visualizing cell-based data doesn't work because writeCellDataReal() expects a valid cell entity with topological relations to nodes to draw the polyhedron.  So rather than trying to write it as a cell-based quantity, the code currently just writes it as a point cloud using writeNodeDataReal().  Does  that no longer work?  I'm pretty sure that trying to use writeCellDataReal() on a logical mesh (which I think is what this change results in) would crash or write junk.

ecoon: Ok, just pushed a commit that adds a test for visualizing logical meshes, this code works fine as is.

ecoon (approved):

PR #868, opened 2024-10-07
Fixes bug in FV discretization

ecoon:
This actually fixes two bugs:

1. The Jacobian term in FV for Dirichlet BCs did not respect the upwinding scheme, so terms were included even if upwinding meant that the kr was not a function of the cell pressure (only a function of the boundary condition pressure).

2. HasComponent("boundary_face") returned False even in cases when HasComponent("face") was true, and therefore the Vandelay map could be used to create a boundary_face on the fly.  This incorrectly resulted in several evaluators not computing on boundary_faces when dependencies were on faces.  I don't believe that these changed answers, but they made for incorrect information to be printed in the Debugger/debug cells.


ecoon:
Closing this -- it will be split into multiple PRs.

PR #870, opened 2024-10-08
Minimal fix for Jacobian bug in FV

ecoon:
FV failed to correctly consider upwinding in the analytic Jacobian term.


ecoon:
Note this is only in the Dirichlet BC case.  If you're upwinding into the domain, then the face rel perm is set by the boundary condition pressure, so the derivative with respect to the (downwind) cell pressure is 0.  This has been around a very long time, and explains poor convergence/timestep we saw in some seepage face problems in overland flow (and in the tpetra test case).

PR #871, opened 2024-10-08
Adds CV::ProvidesComponent() and uses it in vector math

ecoon:
Previously, if A & B are CVs,

A.Update(1, B, 1)

where A.HasComponent("boundary_face") and B.HasComponent("face") would
do nothing, because !B.HasComponent("boundary_face").  This should be
allowed, however, because B.ViewComponent("boundary_face") uses the
Vandelay map and all is fine.

This commit makes the logic:

A.HasComponent(comp) <-- a _writable_ view of comp is available (no change)
A.ProvidesComponent(comp) <-- a _readable_ view of comp is available (new)

and uses ProvidesComponent() where possible to let Update(), Dot(),
etc work through the Vandelay map.

Note also that it is probably implied that:
!(A.HasComponent("boundary_face") && A.HasComponent("face"))
(you can't have both) though I'm not sure whether this breaks anything
or not.

Several changes to code were also made to:

if (A.HasComponent("face")) {
  ...
} else if (A.HasComponent("boundary_face")) {
  ...
}

rather than two independent if conditionals.  Likely this doesn't
really matter, and both are now valid.

### DIFF
--- a/src/data_structures/CompositeVector.cc
+++ b/src/data_structures/CompositeVector.cc
@@ -245,7 +245,7 @@ CompositeVector::operator=(const CompositeVector& other)
       // If both are ghosted, copy the ghosted vector.
       // Exception: boundary_face component created on a fly is not ghosted
       for (name_iterator name = begin(); name != end(); ++name) {
-        bool ghosted = (*name == "boundary_face" && other.ProvidesComponent("face")) ? false : true;
+        bool ghosted = (*name == "boundary_face" && other.HasImportedComponent("face")) ? false : true;
         Teuchos::RCP<Epetra_MultiVector> comp = ViewComponent(*name, ghosted);
         Teuchos::RCP<const Epetra_MultiVector> othercomp = other.ViewComponent(*name, ghosted);
         *comp = *othercomp;
@@ -273,7 +273,7 @@ CompositeVector::operator=(const CompositeVector& other)
 Teuchos::RCP<const Epetra_MultiVector>
 CompositeVector::ViewComponent(const std::string& name, bool ghosted) const
 {
-  AMANZI_ASSERT(ProvidesComponent(name));
+  AMANZI_ASSERT(HasImportedComponent(name));
   if (name == "boundary_face") {
     if (!HasComponent("boundary_face") && HasComponent("face")) {
       ApplyVandelay_();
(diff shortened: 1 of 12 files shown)


ecoon:
Note that the first commit is actually a different PR -- this will be merged after #870 is accepted.

lipnikov:
@ecoon Has vs Provides and writable view vs readable view are confusing terminology. You can write but not read, or you can read but not write, or something else? If we talk about the difference between true component  and derived/vanderlay component, can we use better names.

ecoon:
Happy to change the names if you have suggestions.  You can always read.  What you can't do is use the Vandalay map to import from face --> boundary_face, then write to the boundary_face component (there is no "reverse Vandalay" to export the values form boundary face back to face).

Currently "HasComponent" means that the vector really has that component, so it is writable.  This is our current semantics (no change).

What is new here is "ProvidesComponent" which I meant to hint that the CV can provide this thing to you, but it computes it on the fly and gives you a copy, so you shouldn't expect to be able to write it.  Note that CV::ProvidesComponent is a const function, so you can only call it on const CVs and therefore can only ViewComponent and get a const Epetra_MultiVector.

ecoon:
The important thing to me is that the following doesn't work on master now:

```
CV A, B;
A.AddComponent("boundary_face");
B.AddComponent("face");
.... // set values, etc
A.Update(1, B, -1);
```

On master this currently does nothing, because the implementation of Update() does:

```
for comp in A:
    if B.HasComponent(comp):
        A.ViewComponent(comp).Update(1,B.ViewComponent(comp),-1)
```
and `B.HasComponent("boundary_face")` returns false.

Changing this to:

```
for comp in A:
    if B.ProvidesComponent(comp):
        A.ViewComponent(comp).Update(1,B.ViewComponent(comp),-1)
```

works fine because B is const, so we can get a const view to "boundary_face".

ecoon:
@lipnikov  would "HasDerivedComponent()" be clearer?  I'd rather leave "HasComponent()" alone since it appears throughout the code.

ecoon:
Or maybe `HasImportedComponent()`?  Derived could be read to mean derivatives, but Imported might be clearer?

lipnikov:
Apparently this is not easy: we need to address two cases in one-two words, a true component or imported component. Since I have no good suggestion, feel free to use the original name.

lipnikov (approved):

PR #879, opened 2024-11-15
Correct bad bootstrap argument in ATS CI update

rfiorella:
#877 had the wrong argument in bootstrap, fixing here.


jd-moulton (approved):
Typo fix is pretty clear.

Issue #883, opened 2024-12-17
Inconsistencies in NetCDF configuration in TPLS


rfiorella:
There are still a few inconsistencies in the NetCDF builds that become an issue when Amanzi-ATS is built against other codes that try to find netcdf using nc-config (e.g., E3SM).

For example, in the current containers,
```docker run -it metsi/amanzi-tpls:0.98.9-mpich /home/amanzi_user/install/tpls/bin/nc-config --libs```
yields:
```-L/home/amanzi_user/install/tpls//home/amanzi_user/install/tpls/lib -lnetcdf```

The library path here is incorrect, and this causes an issue compiling ELM-ATS when using `find_package(Amanzi REQUIRED)` as E3SM is already defining a netcdf target, and it is getting the link library path for netcdf from `nc-config --libs`.

The rest of the Amanzi TPLS that import netcdf seem to be getting the library path through different means, and so don't have this issue

ecoon:
Haha, I see this on my local build too (the repeated path).

rfiorella:
A separate issue, but TPLs build fails now at head of ecoon/bugfix_883 due to a problem with superLU download. Not sure if github changed something under the hood recently but github.com/xiaoyeli/superlu/archive doesn't exist anymore (the repo does, just the archive path is missing)

PR #884, opened 2024-12-20
Ecoon/bugfix 883

ecoon:
Removes CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_BINDIR options
These were incorrectly set to $CMAKE_INSTALL_DIR/lib instead of
lib (and bin, respectively), which are the defaults.

This closes https://github.com/amanzi/amanzi/issues/883


rfiorella:
@ecoon Added a commit to try to address #885. 

I'm not sure I've found a satisfying explanation for why these paths changed, and I can't get the new seacas link to work everywhere I've tried it, but curious to know if this TPL issue is experienced by others or just me.

jd-moulton (approved):
These are overdue.  Everyone must have these tarballs in their Downloads directory already or we would hear more complaints.

PR #889, opened 2025-02-13
Changed default floating number to 64

lipnikov:
Metis defined real_t = float. This PR defines real_t = double. Note that MSTK does allocate real_t arrays. Hence, impact is limited to ParMetis.


jd-moulton:
In the places it shows up in MSTK, its defined through the include metis.h and only seems to be used in relation to some io relating to metis and parmetis.   I'll build and test and see if it looks like we are missing anything but this seems benign (not to mention strange that it was hard coded to float - single precision).

jd-moulton (approved):
Looks good to me.   Strange this just doesn't show up in our regular linking or usage patterns.

Issue #864, opened 2024-09-09
Add support for TimeStepControllers in Alquimia_PK


ecoon:
Currently Alquimia_PK handles its own timestep control, e.g.:

https://github.com/amanzi/amanzi/blob/master/src/pks/chemistry/Alquimia_PK.cc#L500
https://github.com/amanzi/amanzi/blob/master/src/pks/chemistry/Alquimia_PK.cc#L985

This was all implemented for general use in TimeStepControllers in the time_integration library.  Alquimia_PK should get refactored to use that library rather than doing it itself.

Issue #888, opened 2025-02-05
MDM_Isotropic double-counts porosity


ecoon:
All other MDM models accept:

```
MDM::mech_dispersion(double t,
                               const AmanziGeometry::Point& xc,
                               const AmanziGeometry::Point& u,
                               int axi_symmetric,
                               double wc,
                               double phi) const;
```

And then they divide u by phi to compute the pore-velocity, and multiply the dispersion by wc (water content).

MDM_Isotropic thinks it is getting saturation, not water content:

```
MDM_Isotropic::mech_dispersion(double t,
                               const AmanziGeometry::Point& xc,
                               const AmanziGeometry::Point& u,
                               int axi_symmetric,
                               double s,
                               double phi) const
```
and multiplies by both s and phi.

Since s is actually passed water content, which already includes the factor of phi, it is double-multiplying phi.  Note that it does NOT divide u by phi.  I'm not sure if this is correct as intended or not.

lipnikov:
@ecoon Isotropic model is not used in Amanzi. For porous media, the Bear model is sometimes referred to as the isotropic model.  Are you suggesting to replace "s * phi" with "wc"?

ecoon:
I've fixed this in ecoon/transport_dev.

Please see changes starting here:  https://github.com/amanzi/amanzi/pull/891/files#diff-58ba1d8bac07c67307575d2755f0cef0a3429e83e6c0c9bb89df15a452f37613L101

Effectively MDM_Isotropic assumes you pass in saturation, what is passed in is actually already phi * s or water content.

Issue #748, opened 2023-05-30
handle errors from the geochemical engine as time-step-cuts?


ecoon:
Discussion from IDEAS call raised the point:

1. Alquimia_PK's current capability for cutting the time step is a bit old, and does not use existing capabilities.
2. Where should this be done? Globally vs. Locally to the cell?  In Alquimia_PK vs Alquimia proper vs geochemistry code vs mpc coupling transport + chemistry?

Issue #746, opened 2023-05-09
Implement temperature in Alquimia PK


ecoon:
Alquimia accepts temperature; Amanzi and ATS can both produce temperature.  Let's hook them up -- this is an easy connection that several have asked about even if they don't necessary have a problem to use it now.

lipnikov:
FYI, native geochemistry uses temperature dependence of reaction rates for almost 2 years. It takes 4-6 lines of the code.

ecoon:
I'm sure that's right.  I'm collecting tickets for people to do -- there has been a surge of interest of reactive transport and my hope is that it is going to culminate in a few other people learning how to develop code.  This is a pretty easy one.  There a bunch more like these in ATS's transport.

PR #891, opened 2025-03-04
WIP: Refactor of (ATS) transport and (Amanzi) Chemistry/Alquimia PKs

ecoon:
This is enabling a major refactor of ATS's reactive transport codebase.

It supports (external) subcycling of reactive+transport with respect to flow, and includes a significant refactor of the chemistry Alquimia_PK to use tags and consistent naming with Alquimia proper. 

This is WIP -- tests for ATS are still in progress, and no effort has been made to compile Amanzi yet, so that will probably not compile or pass tests.  This PR is here to provide a place for discussion around some significant changes.

### DIFF
--- a/src/data_structures/TensorVector.hh
+++ b/src/data_structures/TensorVector.hh
@@ -34,13 +34,13 @@ namespace Amanzi {
 // -----------------------------------------------------------------------------
 struct TensorVector {
   TensorVector(CompositeVectorSpace map_, bool ghosted_ = false)
-    : map(std::move(map_)), ghosted(ghosted_)
+    : map(std::move(map_)), ghosted(ghosted_), rank(-1), dim(-1)
   {
     data.resize(size_());
   }
 
   TensorVector(CompositeVectorSpace map_, int dim_, int rank_, bool ghosted_ = false)
-    : map(std::move(map_)), ghosted(ghosted_)
+    : map(std::move(map_)), ghosted(ghosted_), rank(rank_), dim(dim_)
   {
     data.resize(size_(), WhetStone::Tensor(dim_, rank_));
   }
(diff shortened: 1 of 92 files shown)


ecoon:
@smolins Please take a look at the changes to Alquimia_PK.  The big changes are:

1. supporting failure of a step -- Alquimia_PK can now revert to old state and pick up after a failed step (whether in a single cell or failure of another, coupled PK)
2. Use of time tags to allow sub cycling transport + chem relative to flow.
3. changes names of Amanzi vectors to match Alquimia data
4. respect the differences between "state" (e.g. things that have new and old values and must be stored between steps) and "parameters" (e.g. things that are read-only, and are input to Alquimia but never changed by Alquimia) and "diagnostics" (e.g. things that are out-only, and are just handed to Amanzi for vis).

ecoon:
@smolins A specific question -- I notice that the Alquimia_PK reads a few parameters that are also in the PFLOTRAN card.  For example, Alquimia reads a "aqueous_kinetic_rate_constant" and passes that into Alquimia.  I see that there is also RATE_CONSTANT cards in the corresponding pflotran.in.  

Do you know who should actually be providing these parameters?  I have example runs in ATS (e.g. surface_denitrification) that have it in the card but not in the ATS input file, so a default value of 0 is passed into Alquimia.  

1.

lipnikov:
Too many changes in this MR such as in io.py, etc. Can we split it into two MRs, non-chemistry and chemistry?

ecoon:
This will close #864 #888 #746 #748 #673 #539

ecoon:
> @ecoon Too many changes in this MR such as in io.py, etc. Can we split it into two or more MRs, non-chemistry and chemistry?

I can try to split off as much as possible.  There is a lot of related changes in State, PKs, and transport that cannot be split off.

ecoon:
Note, this will be closed -- it is split into several other PRs.

PR #892, opened 2025-03-10
Moves some helper classes, functions to Amanzi from ATS

ecoon:
This does some general capability additions and quality of life additions to prepare for major changes to Chemistry PK and ATS transport PK.

1. Moves pk_helpers from ATS to Amanzi.  There is significant overlap in ideas with some helper functions in PK_Physical.hh in Amanzi master, and likely these should be merged eventually to help make Amanzi and ATS more consistent.
2. Adds a PK_Physical_Default class from ATS to Amanzi.  This adds more functionality to Amanzi's PK_Physical class, but cannot be the base class for most of Amanzi until Amanzi PKs start to use passwords that are unique, and not empty or "state".  These should be merged, but will require coordination between Amanzi and ATS.
3. Adds the "dot with normal" functionality to Independent variables, moves this capability into a common function to be used across "initial conditions", "independent variables", and others.
4. Adds an "EvaluatorAlias" which is a primary variable evaluator that "aliases" one tag as another tag.  For instance, a PK may subcycle, and so need a "my_next" tag.  At the end of multiple subcycle steps, "my_next" will be the same as the global NEXT tag.  There is no real need to duplicate that data, so an Alias evaluator provides a way of getting NEXT vectors that are actually "my_next" vectors for e.g. vis, checkpointing, observations, temporal interpolation, or other uses.  This is currently only used in subcycling PKs through the `requireEvaluatorAtNext()` helper function.
5. Adds an `EvaluatorIndependentTensorFunction` which computes `TensorVector`s directly rather than requiring a PK to map from a CompositeVector into a tensor.
6. Fixes a bug in `EvaluatorTemporalIntegration` where there was no dependency on time, only on the interpolated fields, which meant things were not correctly updated when only the inner time tag changed.
7.  Does some cleanup in amanzi_xml, mostly keeping comments and making sure it doesn't introduce extra whitespace in weird places.
8. Does some code optimization in src/operators/Limiter*, where vectors were repeatedly reallocated unnecessarily.

### DIFF
--- a/src/data_structures/TensorVector.hh
+++ b/src/data_structures/TensorVector.hh
@@ -34,13 +34,13 @@ namespace Amanzi {
 // -----------------------------------------------------------------------------
 struct TensorVector {
   TensorVector(CompositeVectorSpace map_, bool ghosted_ = false)
-    : map(std::move(map_)), ghosted(ghosted_)
+    : map(std::move(map_)), ghosted(ghosted_), rank(-1), dim(-1)
   {
     data.resize(size_());
   }
 
   TensorVector(CompositeVectorSpace map_, int dim_, int rank_, bool ghosted_ = false)
-    : map(std::move(map_)), ghosted(ghosted_)
+    : map(std::move(map_)), ghosted(ghosted_), rank(rank_), dim(dim_)
   {
     data.resize(size_(), WhetStone::Tensor(dim_, rank_));
   }
@@ -56,9 +56,16 @@ struct TensorVector {
 
   TensorVector& operator=(const TensorVector& other) = default;
 
+  void Zero() {
+    for (auto& t : data) {
+      t.PutScalar(0.);
+    }
+  }
+
   std::vector<WhetStone::Tensor> data;
   CompositeVectorSpace map;
   bool ghosted;
+  int rank, dim;
 
  private:
   int size_()
@@ -82,11 +89,11 @@ class TensorVector_Factory {
   TensorVector_Factory() : d_(0), rank_(0), ghosted_(false){};
 
   int dimension() const { return d_; }
-  const CompositeVectorSpace& map() const { return map_; }
 
   int get_rank() const { return rank_; }
   void set_rank(int rank) { rank_ = rank; }
 
+  const CompositeVectorSpace& map() const { return map_; }
   void set_map(CompositeVectorSpace map)
   {
     map_ = std::move(map);
--- a/src/data_structures/TensorVector.hh
+++ b/src/data_structures/TensorVector.hh
@@ -56,9 +56,9 @@ struct TensorVector {
 
   TensorVector& operator=(const TensorVector& other) = default;
 
-  void Zero() {
+  void PutScalar(double val) {
     for (auto& t : data) {
-      t.PutScalar(0.);
+      t.PutScalar(val);
     }
   }
 
--- a/src/mesh_functions/CompositeVectorFunction.cc
+++ b/src/mesh_functions/CompositeVectorFunction.cc
@@ -21,8 +21,10 @@ namespace Amanzi {
 namespace Functions {
 
 CompositeVectorFunction::CompositeVectorFunction(const Teuchos::RCP<const MeshFunction>& func,
-                                                 const std::vector<std::string>& names)
-  : func_(func)
+        const std::vector<std::string>& names,
+        bool dot_with_normal)
+  : func_(func),
+    dot_with_normal_(dot_with_normal)
 {
   AMANZI_ASSERT(names.size() == func->size());
 
@@ -36,6 +38,16 @@ CompositeVectorFunction::CompositeVectorFunction(const Teuchos::RCP<const MeshFu
 
 void
 CompositeVectorFunction::Compute(double time,
+                                 const Teuchos::Ptr<CompositeVector>& cv,
+                                 const VerboseObject* vo)
+{
+  if (dot_with_normal_) ComputeDotWithNormal_(time, cv, vo);
+  else Compute_(time, cv, vo);
+}
+
+
+void
+CompositeVectorFunction::Compute_(double time,
                                  const Teuchos::Ptr<CompositeVector>& cv,
                                  const VerboseObject* vo)
 {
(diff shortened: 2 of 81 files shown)


src/data_structures/TensorVector.hh
```diff
@@ -56,9 +56,16 @@ struct TensorVector {
 
   TensorVector& operator=(const TensorVector& other) = default;
 
+  void Zero() {
```
lipnikov, line 59: Do we really need a new name in addition to PutScalar(val) which is used in many other places?

src/mesh_functions/CompositeVectorFunction.cc
```diff
@@ -168,5 +180,78 @@ CompositeVectorFunction::Compute(double time,
...
+
+        for (int i = 0; i < ndofs; ++i) {
+          const AmanziGeometry::Point& normal = cv->Mesh()->getFaceNormal(f, cells[i], &dir);
+          dat_f[0][g + i] = (vel * normal) * dir;
+        }
+      }
+    }
+  }
+
+  // // check divergence free
```
lipnikov, line 237: Remove  debug code

src/operators/FCT.cc
```diff
@@ -79,11 +79,13 @@ FCT::Compute(const CompositeVector& flux_lo,
   const auto& bc_value = bc.bc_value();
 
   Teuchos::RCP<const CompositeVector> bounds;
-  if (limiter_->get_external_bounds())
+  if (limiter_->get_external_bounds()) {
     bounds = limiter_->get_bounds();
-  else
-    bounds =
-      limiter_->BoundsForCells(*field_, bc_model, bc_value, OPERATOR_LIMITER_STENCIL_C2C_ALL);
+  } else {
```
lipnikov, line 84: I do not understand the new logic, bounds are re-computed anyway.

src/operators/LimiterCell.cc
```diff
@@ -249,6 +257,8 @@ LimiterCell::ApplyLimiter(const AmanziMesh::cEntity_ID_View& ids,
     limiter_->Scale(cfl_);
     data_->Scale(cfl_);
   }
+
+  lifting->data()->ScatterMasterToGhosted("cell");
```
lipnikov, line 261: limiter was created using owned cells, so this call is not needed

src/pks/PK_Helpers.cc
```diff
@@ -0,0 +1,288 @@
...
+  } else if (bcs.bc_model()[f] == Operators::OPERATOR_BC_DIRICHLET) {
+    return bcs.bc_value()[f];
+    // } else if (u.HasComponent("boundary_face")) {
+    //   AmanziMesh::Entity_ID bf = getFaceOnBoundaryBoundaryFace(*u.Mesh(), f);
+    //   return (*u.ViewComponent("boundary_face",false))[0][bf];
+  } else {
+    auto c = getFaceOnBoundaryInternalCell(*u.Mesh(), f);
+    return (*u.ViewComponent("cell", false))[0][c];
+  }
+  return -1;
```
lipnikov, line 80: If we are here, we should abort instead of returning a meaningful value.

src/pks/PK_Helpers.cc
```diff
@@ -0,0 +1,288 @@
...
+}
+
+
+// -----------------------------------------------------------------------------
+// Get the directional int for a face that is on the boundary.
+// -----------------------------------------------------------------------------
+int
+getBoundaryDirection(const AmanziMesh::Mesh& mesh, AmanziMesh::Entity_ID f)
```
lipnikov, line 88: Probably, this should be moved to MeshHelpers_impl.hh

src/pks/PK_Helpers.cc
```diff
@@ -0,0 +1,288 @@
...
+      new_positions[n] = Amanzi::AmanziGeometry::Point{ nodes[0][n], nodes[1][n] };
+    } else {
+      new_positions[n] = Amanzi::AmanziGeometry::Point{ nodes[0][n], nodes[1][n], nodes[2][n] };
+    }
+  }
+  Amanzi::AmanziMesh::deform(mesh, node_ids, new_positions);
+}
+
+int
+commMaxValLoc(const Comm_type& comm, const ValLoc& local, ValLoc& global)
```
lipnikov, line 265: Since this is a delegating function, maybe it has to be named after the MPI function, like allReduceMax() or similar.

src/pks/PK_Helpers.cc
```diff
@@ -0,0 +1,288 @@
...
+{
+  ValLoc local{ 0., 0 };
+  for (int i = 0; i != vec.MyLength(); ++i) {
+    if (vec[i] > local.value) {
+      local.value = vec[i];
+      local.gid = vec.Map().GID(i);
+    }
+  }
+  ValLoc global{ 0., 0 };
+  int ierr = commMaxValLoc(vec.Comm(), local, global);
```
lipnikov, line 283: Do we need such a code granularity? Do we need this specialized function?

src/data_structures/TensorVector.hh
ecoon: Changing Zero() to PutScalar(0.) in a few places; agreed there is no need for different names.

ecoon: In doing this, I found a new data structure -- src/common/chemistry/utility/MatrixBlock.  Is there a reason this exists? We have WhetStone::Tensor, TensorVector, and other things that I would assume duplicate this capability.

src/pks/PK_Helpers.cc
ecoon: Effectively this replaces Teuchos::reduceAll, which doesn't support MAX_LOC reductions.  I agree the name is bad.  In the tpetra branch, we put this in src/utils/Reductions.hh -- I'll do the same here.

ecoon: Actually, I'm surprised Amanzi doesn't use something like -- do you not report the location/cell of the maximum error norm?

src/pks/PK_Helpers.cc
ecoon: agreed, I will change this to the same as our tpetra version (this predates that), which is `Reductions::reduceAllLoc()`

src/operators/FCT.cc
ecoon: So it's possible I missed the "external bounds" parts of this.  I wanted to avoid this:

https://github.com/amanzi/amanzi/blob/master/src/operators/LimiterCell.cc#L820

at every step of transport.

Currently, on master, if I just set a breakpoint in LimiterCell.cc at all allocations of new bounds vectors, it creates a new vector 4 times per subcycled step in Amanzi's transport test transport_2D_long.  After this PR, it creates a new vector twice in the entire run, then reuses the bounds vector instead of recreating it.

lipnikov: If you change a field, you may want to update bounds to stay monotone. It is fine to allocate it once, but my understanding that you do not update it during subcycling, b/c it already exists. Do you agree?

src/pks/PK_Helpers.cc
lipnikov: If you commit more capabilities to Amanzi, we will certainly use them ;-)

src/data_structures/TensorVector.hh
lipnikov: Chemistry's CMake is processed before WhetStone's. I agree, but do not want to think now about potential differences in APIs.

src/pks/PK_Helpers.hh
```diff
@@ -0,0 +1,142 @@
...
+getFaceOnBoundaryValue(AmanziMesh::Entity_ID f,
+                       const CompositeVector& u,
+                       const Operators::BCs& bcs);
+
+
+// -----------------------------------------------------------------------------
+// Get the directional int for a face that is on the boundary.
+// -----------------------------------------------------------------------------
+int
+getBoundaryDirection(const AmanziMesh::Mesh& mesh, AmanziMesh::Entity_ID f);
```
lipnikov, line 51: I think the function was relocated.

src/pks/PK_Physical_Default.cc
```diff
@@ -0,0 +1,116 @@
...
+PK_Physical_Default::Setup()
+{
+  // set up the debugger
+  Teuchos::RCP<Teuchos::ParameterList> vo_plist = plist_;
+  if (plist_->isSublist(name_ + " verbose object")) {
+    vo_plist = Teuchos::rcp(new Teuchos::ParameterList(*plist_));
+    vo_plist->set("verbose object", plist_->sublist(name_ + " verbose object"));
+  }
+
+  db_ = Teuchos::rcp(new Debugger(mesh_, name_, *vo_plist));
```
lipnikov, line 46: Looks like debugger is the mandatory variable. Is it really the case?

src/utils/TimeStepManager.cc
```diff
@@ -56,7 +57,7 @@ TimeStepManager::TimeStepManager(Teuchos::ParameterList& plist)
     manual_override_ = true;
   }
 
-  vo_ = Teuchos::rcp(new VerboseObject("TimeStepManager", plist));
+  vo_ = Teuchos::rcp(new VerboseObject(Keys::cleanPListName(plist), plist));
```
lipnikov, line 60: This leads to "ANONYMOUS" prefix in some Amanzi tests. Can we support/provide a default name (TimeStepManager for now), so everybody be happy?

lipnikov (approved):
Approval is contingent to resolving the raised questions/comments.

src/pks/PK_Physical_Default.cc
ecoon: This saves from having to do a "if db_ != Teuchos::null" everywhere in the code.  If you provide no debug cells, then the debugger won't do anything, and won't be a significant performance overhead.

I don't think it requires anything in the ParameterList, so it's not like the code errors if "debug cells" aren't provided.

So yes, db is always constructed.  But if you don't provide "debug cells" it doesn't do anything.

PR #893, opened 2025-03-12
reformatting all xml

ecoon:
Future changes to Chemistry will require changes to a significant number of xml files.  These changes will be done through a script that uses amanzi_xml.

However, currently most xml files are formatted differently from how amanzi_xml formats.  This means that after the above changes are made, there will be many many formatting differences that hide the actual changes.

This PR does two things:

1. It changes amanzi_xml's formatting to act as much like Teuchos::XMLObject's writers as possible.
2. It runs a "null op" read and write of all Amanzi xml files.  These make no actual changes, but do reformat things to a standard formatting approach, e.g. same whitespace conventions and same scientific notation on floats.

### DIFF
--- a/demos/phase1/f-area/chemistry/1d-uo2-17-component.xml
+++ b/demos/phase1/f-area/chemistry/1d-uo2-17-component.xml
@@ -1,6 +1,4 @@
-<!-- 1D 5 component UO2 reactive transport problem for comparison to pflotran -->
 <ParameterList name="Main">
-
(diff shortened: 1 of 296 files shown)


ecoon:
@lipnikov  let me see if I can do that.  I'm not manually writing these, they are written by the python ElementTree library, but I can see if I can figure out a way to reformat.  I agree with preferring 1.0 vs 1 for floats, it reinforces the type.

Issue #316, opened 2019-07-03
Rename local_matrices to local_op


lipnikov:
Teuchos::RCP<Op> local_matrices() { return local_op_; } violates our naming convention. In addition, access to matrices requires to call "matrices" twice which reduces code readability, e.g. diff->local_matrices()->matrices. I suggest to rename this function to local_op().

Issue #315, opened 2019-06-23
Add boundary map for nodes


lipnikov:
Add support of a boundary map for mesh nodes. Besides its usage in various boundary conditions, this map will be used to deform structured mesh in unit tests, e.g. for convergence analysis.

PR #894, opened 2025-03-18
Major refactor of Chemistry_PK, Alquimia_PK

ecoon:
- Chemistry_PK was largely unused, and now does only a few specific things, most notably timestep size control through TimestepController
- Alquimia_PK now uses tags
- Alquimia_PK now has a struct for all vectors to amortize State calls
- Alquimia_PK names/keys now use variable names that match Alquimia, not Amanzi chemistry
- runs tools/amanzi_xml/fix_chemistry_ts_control.py on all xml, which changes xml parameters to use "timestep controller" object

### DIFF
--- a/demos/phase3/tank/Tank-alq_v2.xml
+++ b/demos/phase3/tank/Tank-alq_v2.xml
@@ -43,9 +43,9 @@
       <density>998.2</density>
       <dissolved_components>
 	<solutes>
-	  <solute coefficient_of_diffusion="5.e-12" first_order_decay_constant="0">Tracer</solute>
-	  <solute coefficient_of_diffusion="5.e-12" first_order_decay_constant="0">Sr90</solute>
-	  <solute coefficient_of_diffusion="5.e-12" first_order_decay_constant="0">Cs137</solute>
+	  <solute coefficient_of_diffusion="5.e-12" first_order_decay_rate_constant="0">Tracer</solute>
+	  <solute coefficient_of_diffusion="5.e-12" first_order_decay_rate_constant="0">Sr90</solute>
+	  <solute coefficient_of_diffusion="5.e-12" first_order_decay_rate_constant="0">Cs137</solute>
 	</solutes>
       </dissolved_components>
     </liquid_phase>
--- a/doc/input_spec/AmanziInputSpec-s.rst
+++ b/doc/input_spec/AmanziInputSpec-s.rst
@@ -990,7 +990,7 @@ The subelement ``primaries`` is used for specifying reactive and non-reactive pr
 
     * ``coefficient_of_diffusion`` = "exponential", this is an optional attribute
 
-    * ``first_order_decay_constant`` = "exponential", this is an optional attribute
+    * ``first_order_decay_rate_constant`` = "exponential", this is an optional attribute
 
     * ``forward_rate`` = "exponential", this is a required attribute when being used with non-reactive primaries/solutes and automatically generating the chemistry engine input file
 
--- a/doc/input_spec/AmanziInputSpec-u.rst
+++ b/doc/input_spec/AmanziInputSpec-u.rst
@@ -1477,7 +1477,7 @@ The subelement ``primaries`` is used for specifying reactive and non-reactive pr
 
     * ``coefficient_of_diffusion`` = "double", this is an optional attribute
 
-    * ``first_order_decay_constant`` = "double", this is an optional attribute
+    * ``first_order_decay_rate_constant`` = "double", this is an optional attribute
 
     * ``forward_rate`` = "double", this is a required attribute when being used with non-reactive primaries/solutes and automatically generating the chemistry engine input file
 
--- a/doc/input_spec/schema/amanzi.xsd
+++ b/doc/input_spec/schema/amanzi.xsd
@@ -1591,7 +1591,7 @@ TODO Can we do validation on known names plus allow unrecognised ones.  Well I t
 	<xs:simpleContent>
 	<xs:extension base="xs:string">
 		<xs:attribute name="coefficient_of_diffusion" type="typedef_exponential"/>
-		<xs:attribute name="first_order_decay_constant" type="typedef_exponential"/>	
+		<xs:attribute name="first_order_decay_rate_constant" type="typedef_exponential"/>	
 		<xs:attribute name="forward_rate" type="typedef_exponential" />
 		<xs:attribute name="backward_rate" type="typedef_exponential" />
 		<xs:attribute name="molar_mass" type="typedef_exponential" />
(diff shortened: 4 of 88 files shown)


ecoon:
This will close #864

lipnikov:
Chemistry tests in the user guide are failing. Fixing may take time... A few errors:

"reduction_factor_ <= 1.0" failed in file: [...]TimestepControllerStandard.cc 

initializing eval: "primary_activity_[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range

Assertion: "ghosted_" failed in file: [...]CompositeVector.cc, at line: 348

lipnikov:
@ecoon amanzi-u-1d-tracer-alq-crunch.xml works for master and does not work for this branch. Please look at this test.

ecoon:
The ghosted assertions should pass now with the new commit.  

```"reduction_factor_ <= 1.0" failed in file: [...]TimestepControllerStandard.cc```

Why would one provide a reduction factor > 1?  That's not a reduction?  This sounds like an incorrect input file.

How do I run the user guide tests?  bootstrap with --enable-reg_tests doesn't seem to run amanzi-u-1d-tracer-alq-crunch.xml.

**Comment by ecoon** [comment]
Ok, so that is the problem.  In Chemistry_PK, dt_cut_factor divides dt, while in TimestepManager dt_reduction_factor multiplies the dt.  

Ok, I'll fix this as well, keeping the non-native spec the same in Amanzi.  However, @lipnikov  you may want to look at this and choose what you would like Amanzi's input spec to do -- currently it is not consistent in that `InputCoverterU::TranslateTimeIntegrator_` reads dt_cut_factor as a multiplier (e.g. < 1) while `InputConverterU::TranslateChemistry_` reads time_step_cut_factor as a divisor (e.g. > 1).

**Comment by lipnikov** [comment]
the easiest way is to use files in e.g. test_suite/benchmarking/chemistry/farea_1d to run the test. Alquimia test is giving me a segfault.

**Comment by ecoon** [comment]
Ok, so these tests aren't run with --enable-reg_tests.  I'm not sure if they're supposed to be or not.  But I'll look at these.

Also, the reg_tests don't pass for me on master -- the last commit 15cfde8 was required to get them to run on master (otherwise they seg fault during teardown/destruction).  But even with that commit, I now see some failing tests on master during the comparison stage:

e.g. compare_butler_pod_2d_observation


```
❯ VERBOSE=1 ctest -R compare_butler_pod --output-on-failure
Test project /home/ecoon/code/ats/amanzi/build/ecoon-pre_chemistry_transport_dev/ubuntu-24.04-epetra/debug
    Start 315: run_butler_pod_2d_for_comparison
1/2 Test #315: run_butler_pod_2d_for_comparison ....   Passed   20.42 sec
    Start 316: compare_butler_pod_2d_observation
2/2 Test #316: compare_butler_pod_2d_observation ...***Failed  Required regular expression not found. Regex=[Comparison Passed
]  0.12 sec
Comparison Failed for observation drawdown:
  error 0.00567034 > tolerance 0.002 in region Obs1
  error 0.00513786 > tolerance 0.002 in region Obs2

```

**Comment by lipnikov** [comment]
@ecoon I  am confused about changes to the time step manager on October 17. The new code suggest 3.64 sec instead of 1.5 sec

```
1: CycleDriver        |  Cycle 0: time = 0 s, dt = 1 s
...
1: TimeStepManager    |   Accepted proposed dT=1.5 s
1: CycleDriver        |  New time = 1 s

AFTER: beb928213f2d72ec32ceff6c4c77d1589aa07a21

1: CycleDriver        |  Cycle 0: time = 0 s, dt = 1 s
...
1: TimeStepManager    |   Proposed dT=1.5 s, stability from previous step changes it to 3.64 s
1: TimeStepManager    |   Proposed dT=3.64 s, events limit it to 2 s
1: CycleDriver        |  New time = 1 s
```

ecoon:
Yes, I was noticing that InputConverterU_Chemistry.cc:397 sets the min timestep size at 1e10.  In some of these tests (butler_pod_2d) it seems that "min_time_step" isn't set in the (non-native) input file, and so the native spec gets "min timestep [s]" = 1e10, which crashes in TimeStepManager.  I guess the semantics of the old dt control in Chemistry ignored this?

Do you want to change that default value?  It seems odd to me that the min dt default would be the same as the max dt default.  In TimeStepManager, that default is 1e-10.

ecoon:
@lipnikov  Actually I don't really like these defaults at all.  There are currently two sets of default parameters for unstructured chemistry time stepping:

1. The TimestepManager sets sane defaults for its parameters
2. The InputConverterU_Chemistry overrides those defaults with its own defaults.

I would suggest setting the default value in InputConverterU_Chemistry to be NAN, and only writing the parameter in the native file if it is specifically listed in the non-native file.   This allows the TimestepController to set sane defaults.

Also, by default, if no timestep parameters are provided (see, for instance, amanzi-s-1d-surface-complexation-alq-pflo.xml ), then "fixed" is used, and the fixed step size defaults to 1.e7, which is totally random.

Instead, I would suggest setting the default to "standard" which will adapt the timestep to the iterations, and, more importantly, set the chemistry default step size to "very large" which allows transport to control the timestep, and only reduce if if chemistry fails to converge.

lipnikov:
@ecoon  I have no objections as long as the user guide tests are accurate and fast as now. The Newton method typically requires smaller values (4 and 8) (compared to NKA) to stay inside a convergence region. This is my understanding why Alquimia developers suggested such numbers. At the end, my major concern is the operator split error between transport and chemistry PKs.

ecoon:
> @ecoon I have no objections as long as the user guide tests are accurate and fast as now. The Newton method typically requires smaller values (4 and 8) (compared to NKA) to stay inside a convergence region. This is my understanding why Alquimia developers suggested such numbers. At the end, my major concern is the operator split error between transport and chemistry PKs.

That's fair, and I understand the lack of control on splitting error.  I did some checks on the user guide test for surface_complexation -- the input file, unaltered, takes 1e4 timestep size in chemistry.  This must have been ignored on master, because the transport step is 1e7, so without changing the chemistry step, it would take ~1000 steps.  The increase limit is 4 iterations, and the step size never increased because it always took 4 iterations.  However, if I changed the chemistry step size to 1e7, it also converged with 4 iterations.  Obviously, I don't know what is going on with splitting error, but this is problematic.  

For now and for this PR, I'm just changing the user guide input files to get reasonable timestep sizes, and ensuring the solution is the same (by the eyeball norm).

ecoon:
@lipnikov Any remaining concerns?  I think this is ready now.

lipnikov (approved):
Besides old issues with chemistry tests (Alqumia vs Native), it is good to go

Issue #760, opened 2023-07-18
Update Docker support and GitHub actions for multiple architectures


jd-moulton:
The newer Macs use M1 (or M2) chips, which are ARM64 based.  Our current images assume AMD64 (essentially intel x86_64), and performance through emulation on new Macs is terrible (as expected).   Fortunately, Docker has a nice multiarch solution that has been built in to the buildx tool.   Although at somepoint soon we want to integrate our TPL builds into CI, for now we'll assume these are done manually.  This assumption leaves us with two critical updates:

- [x] switch over to buildx for TPLs ( #761 )
- [x] update GitHub actions to use buildx as well ( #762 )

rfiorella:
Closing, superseded by #896.

Issue #761, opened 2023-07-18
Update Docker-TPLs to use buildx and support multiple architectures


jd-moulton:
To support the new Macs with M1 (or M2) chips based on ARM64 we need to add multiarch support to our Docker files.  It looks like the cleanest solution is to use features of the buildx tool.

rfiorella:
Closing, superseded by #896

PR #896, opened 2025-03-24
Enable arm64 runner and produce multi-arch CI image

### LINKED ISSUE #762: Update GitHub actions to support multiple architectures
GitHub actions need to support builds for M1/M2 ARM64 based systems, as well as the more common AMD64 (i.e., x86_64).  Part of the solution is using buildx, but other parts of our workflow may be updated as github actions are maturing.   Pushing to dockerhub may be an issue, so this may need to stay explicit.

rfiorella:
Enables the new arm64 runners for CI jobs, runs the build and test jobs on x86_64 and arm architectures, and then modifies the docker manifest in order to produce a multi architecture image on DockerHub. 

One limitation is that the current workflow produces 3 tags on docker hub. It may be possible to simplify with CI artifacts, but the artifacts might be large enough that it might be preferable to have multiple tags rather than caching docker layers.

Closes #762.


rfiorella:
@jd-moulton seems like the TPL images build pretty quickly on CI (about an hour on each architecture). Conflict is due to dummy file added to master. I think based on our conversation earlier this one is ready to merge now.

rfiorella:
With the TPL updates made, this PR would also close #760 and #761

PR #897, opened 2025-04-09
Transport changes to enable ATS transport refactor

ecoon:
- changes "parameters for X" in MDM factory to "X parameters" to match all other Amanzi/ATS factories
- fixes bug in MDM_Scalar whereby water content and saturation were incorrectly used
- Adds an MDM Evaluator
- improves documentation and increases options for parameter names in MDM models, e.g. "dispersion coefficient" vs "dispersivity," with correct units


ecoon:
closes #888

PR #900, opened 2025-05-04
Updating TPLs to version 0.98.11 - upgrades zlib and patches for CMake 4 portability

jd-moulton:
Issues have emerged with newer compilers (Apple Clang 17, and gcc 14), newer versions of the international character support library (libicu) and on RedHat/CentOS that we address in this minor update of the TPLs.

- [x] zlib upgraded to 1.3.1 to resolve build problems with Apple Clang 17 (@fmyuan)
- [x] Pulls in a patch for kokkos-kernels from Trilinos developers (@fmyuan)
- [x]  libicu versions later than 0.76 need C++17 to process headers (@jd-moulton)
- [x] Revisits removing the build and library directories for netcdf as this breaks seacas builds on CentOS due to use of both lib and lib64 directories.
- [x] Revisits patch of parmetis.h to set float to double that breaks Trilinos builds for gcc 14. 
- [ ] CMake 4.0 sets a new minimum version of 3.5 (but we have TPLs still allowing 2.8). 
   - [x] Initial fix using -DCMAKE_POLICY_VERSION_MINIMUM [5e73f4f](https://github.com/amanzi/amanzi/pull/900/commits/5e73f4fc0aeecf6780e1bcacd089fa501c39c883) works well for the CMake part, 
   - [ ] But a possible issue with libicu has shown up on older OSX, testing more.
- [x] There does not seem to be a problem with python > 3.12, so not pulling in PR from @rfiorella 
- [ ] Looks like the bootstrap is building TPLs in serial despite --parallel=8 option for @lipnikov.  This appears to be limited to OSX and some XCode/brew setups. 
- [x] Patch step for HDF5 has failed. File to be patched does not exist.
- [x] NetCFD-Fortran is not patched cleanly.
- [x] MOAB 
  - [x] CMake 4 issue: Configuration of MOAB is failing (Should we comment out line 167 in root CMaleList.txt ?) MOAB uses incorrect spelling of TestEndianness.c (single 'n').  Not sure the cleanest way to fix this.
  - [x] CMake 3(recent) and 4 and all gcc versions: Linking of MOAB fails due to  -lMPI_LIBRARY-NOTFOUND string; however, MPI_LIBRARY is correct in CMakeCache.txt. MPI_LIBRARY is destroyed during call of find_package( HDF5_MOAB Required). A workaround is to make a temporary copy of MPI_LIBRARY and to reset it after. 
- [ ] PETSc fails due to "ImportError: cannot import name '_parse_makefile' from 'sysconfig'" (in petsc-3.20-source/config/gmakegen.py). A workaround is to import from sysconfig.__main__ which I find strange (might python 3.13.5).
- [x] SEACAS build fails due conflict between two definitions of NC_SZIP_NN ( seacas-2022-02-16-source/packages/seacas/libraries/exodus/src/ex_utils.c, line 1745 and TPLs-0.98.11-clang-17.0.0-ompi-5.0.7-shared/include/netcdf.h, line 2129).
- [x] Alquimia configuration is failing due to cmake_minimum_version set to 2.8
- [x] Potential bug in Alquimia, in alquimia/pflotran_alquimia_interface.F90, line 1005. This unit is not matching what is in line 1014.
- [x] cut-and-paste from #882 
 
Docker images of the TPLs are now built through CI so updates here will be tested automatically.


fmyuan:
I tested this branch on my Mac (M3 Chip), with clang-17 and Openmpi-4.1.8. It's good.

lipnikov:
How about
-DCMAKE_POLICY_VERSION_MINIMUM=3.5

ecoon (approved):

smolins:
I would like to support the merging of commit 33b178c into master. LBNL institutional clusters run Rocky Linux and the current set up for seacas and netcdf results in errors.

jd-moulton:
@smolins It's all "fixed" in the commit after the one mentioned by Rich [b185d6b](https://github.com/amanzi/amanzi/pull/900/commits/b185d6b943f6ee88dd68582fd3aad0073ddcf2e2).  So this PR should have it right and I have built it on RedHat.  Can you try this PR, I plan to merge it this week.

lipnikov (changes requested):
A few problems were identified, see new bullets.

jd-moulton:
Newer gcc compilers seem to catch issues that easily pass through older (gcc < 14) compilers.  Default on our current CI Ubuntu systems is gcc 11, so @rfiorella is looking at adding an option to use images with gcc 14 (or newer).

lipnikov:
@jd-moulton @rfiorella I'll delay testing of User Guide and Tutorial until SEACAS and NetCFD-c be updated, so that I'll be able to rebuild my TPLs.

rfiorella:
Finished a large set of changes that address a few of the items raised by @lipnikov: 
- the HDF5 file being patched has moved from config/cmake_ext_mod/ to config/cmake. I suspect a new patch will need to be made still, but don't know enough about the origin of the patch to know if it's still required. Also looks like it affects only MacOS and I haven't yet tested there yet.
- the NetCDF-Fortran patch that is causing an issue has been fixed in the TPL as part of the version bump, so this patch has been removed.
- the deploy-tpls-docker.sh script and CI jobs have been updated to make it easier to test on gcc-11, gcc-14, and gcc-latest (currently 15.1). The gcc-14 and gcc-latest are Debian-based images though, so there are some additional slight differences. (though since they both use apt, the changes are actually quite minor).
- upgrading to SEACAS v2025-04-14 resolves the double-definition issue.

but unfortunately, these changes have revealed a few new issues with newer compilers:
- [ ] upgrade of SEACAS breaks MSTK for gcc-14: `Could NOT find ExodusII (missing: ExodusII_LIBRARY ExodusII_INCLUDE_DIR)`
- [ ] superLU on gcc-latest has a very similar issue to that previously described for SEACAS v2022-xx-xx (multiple definitions of the same variable), but, the version of superLU we are using has not been updated since 2021 and 0.97.9. <- Fixed by upgrading to 7.0.1.
- [ ] superLU_dist on gcc-latest has conflicting type compile error.

These are still on the rich/amanzi-tpls-0.98.11 branch since there are some WIP patches. I will cherry-pick when these are fixed.

rfiorella:
Several items in the checklist addressed by #907

fmyuan (approved):

PR #903, opened 2025-05-17
Refactory of unstructured converter

lipnikov:
More genertatily of coupling PKs into MPC PKs requires some rules. This PR removes hard-coded rules and replaces them with generic code.

Removed files using unsupported XML format.
Removed deprecated format for <process_kernels> and upgraded exisitng XML files. Changed verbosity level in the chemistry PK for some output. Changed rules for coupling PKs, physics-based coupling (flow + energy) was moved up. Removed debug output, fixed recently introduced conversion bugs. Upgrated XML files in the User Guide.


src/pks/chemistry/Chemistry_PK.cc
```diff
@@ -96,7 +96,7 @@ Chemistry_PK::AdvanceStep(double t_old, double t_new, bool reinit)
   double dt = t_new - t_old;
 
   Teuchos::OSTab tab = vo_->getOSTab();
-  if (vo_->os_OK(Teuchos::VERB_LOW))
+  if (vo_->os_OK(Teuchos::VERB_EXTREME))
```
ecoon, line 99: We'll have to come up with as strategy for log file verbosity levels if there is going to be shared PK code.  Currently this is expected at VERB_LOW for ATS, though I think I may end up making it VERB_MEDIUM.  But definitely not VERB_EXTREME.

PR #912, opened 2025-07-16
Update Amanzi CI to run across multiple OS, GCC, MPI, CMake

rfiorella:
Makes changes to the Amanzi CI matrix so that tests run on more architectures and MPI versions to
hopefully reveal issues with different MPI, CMake, and compilers are spotted faster.


rfiorella:
The failing test looks like it arises from an issue newer MPICH has with ParallelDescriptor::EndParallel from BoxLib. To pass CI at this point we'll need to do one of the following:
1) add a new patch to CCSE
2) migrate to AMReX from CCSE
3) disable structured for now

rfiorella:
Split into #917 and later PR TBD.

PR #915, opened 2025-07-23
clang-format 19 with new file

ecoon:
clang-format 19 changes the input file from the last version (16?)

I've tried to mimic as best as possible.

One thing I don't like is that clang-format still does not allow 

```
if (condition) {
  do_thing();
}
```

to be treated differently from:

```
if (condition) do_thing();
```

Both are either single-line or not single-line in all cases.  I would prefer to single-line the second and not the first.  But that isn't possible, so I've gone with not single-line them all.  As a result, the second is formatted as:

```
if (condition)
  do_thing();
```
in all cases.

But I think this is better than the alternative, because we want the first to be multi-line to allow for gdb break statements inside of the conditional.


ecoon:
> I know this is really for developers, so can I skip it for this release, or is it something you want in there.

I want this in there.  To me this is the last thing we do before release.

ecoon:
Suggested changes:

BreakAfterReturnType == TopLevelDefinitions
https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html#breakafterreturntype

AlignOperands == AlignAfterOperator
https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html#alignoperands

lipnikov (approved):

lipnikov:
FYI: all tests passed for me.

ecoon:
Latest pushes:

1. make the above changes on BreakAfterReturnType, AlignOperands, and BreakBeforeTernary.
2. add a python script that enforces the rule of: if braces are included, do not make if or for loops 1-liners.  If braces aren't included, make them 1-liners if they fit on one line.
3. Updated clang-format.sh script to call the python script
4. run on all source

PR #918, opened 2025-07-25
Converter refactory with a cleaner history

lipnikov:
More genertatily of coupling PKs into MPC PKs requires some rules. This PR removes hard-coded rules and replaces them with generic code.

Removed files using unsupported XML format.
Removed deprecated format for <process_kernels> and upgraded exisitng XML files. Changed verbosity level in the chemistry PK for some output. Changed rules for coupling PKs, physics-based coupling (flow + energy) was moved up. Removed debug output, fixed recently introduced conversion bugs. Upgrated XML files in the User Guide.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -135,7 +135,7 @@ no_color=$FALSE
 
 # Amanzi build configuration
 # -- components
-structured=$TRUE
+structured=$FALSE
 unstructured=$TRUE
 geochemistry=$TRUE
 amanzi_physics=${TRUE}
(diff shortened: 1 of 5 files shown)


lipnikov:
@fmyuan This header is related to Amanzi 1.7-dev. The issue is due to different include lists for ATS physics and Amanzi physics. If you compile ATS, use Amanzi release 1.6. Meanwhile, I'll analyze if this header can be relocated to be visible by ATS.

fmyuan:
> @fmyuan This header is related to Amanzi 1.7-dev. The issue is due to different include lists for ATS physics and Amanzi physics. If you compile ATS, use Amanzi release 1.6. Meanwhile, I'll analyze if this header can be relocated to be visible by ATS.

That's what I thought. It's strange to me that other 4 includes of same *.hh, e.g. pks/flow/Flow_pk.hh, are OK, but that in transport pk has issue.

PR #930, opened 2025-09-18
[WIP] update tpls to 0.98.13

rfiorella:
TPL updates:
- hypre 2.29->2.31
- PETSc 3.20->3.21
- fmt 11.2.0 (new TPL of SEACAS)
- SEACAS v2022-02-16 -> v2025-04-14
- superLU 5.2.2 -> 6.0.1
- superLU_dist 6.4.0 -> 8.2.1
- netcdf-c 4.9.2 -> 4.9.3

Patch updates:
- [ ] new patches for superLU and superLU_dist
- [ ] regenerate HDF5 patch for 1.14.6
- [ ] revised PETSc patches
- [ ] fix failing silo patches


lipnikov (changes requested):
MOAB build fails:
src/io/ReadHDF5Dataset.cpp:108:9: error: use of undeclared identifier 'H5Pset_dxpl_mpio'

Reason: homebrew installation is used instead of TPL installation.

rfiorella:
@lipnikov just to clarify (and apologies I think I asked this earlier this summer, MOAB build fails linking to homebrew hdf5?

Is this homebrew hdf5 or hdf5-mpi?

rfiorella:
Ah, so it's linking to homebrew hdf5 when it should be linking that from the TPL build?

lipnikov:
No, compiler uses headers from a wrong directory.

rfiorella:
Oh right. Wasn't thinking through the implications of the fix you suggested.

config/SuperBuild/templates/silo-4.11.1-h5epr-semi-colon.patch
```diff
@@ -0,0 +1,190 @@
...
+-    return -1; 
++    return -1;
+ }
+ 
+ static int find_block_to_preempt(H5FD_silo_t *file)
+@@ -829,7 +835,7 @@
+     block = &(file->block_list[blidx]);
+ 
+     HDassert(block->buf);
+-
```
lipnikov, line 56: Majority of this patch are spaces. Fortunately, this is not the package we update often.

lipnikov:
@rfiorella Any thoughts about PETSc? Looks like the recent PETSc did not fix this issue.

config/SuperBuild/templates/silo-4.11.1-h5epr-semi-colon.patch
rfiorella: Huh, this picked up too many changes, must have been a bad setting on my editor. Should have just been the H5EPR_SEMI_COLON lines. The 4.11 patch seemed was failing due to being unable to match hunks 2-4. I'll clean up this patch so that it just includes what's necessary.

rfiorella:
There's this patch in a PETSc google group, looks like the same fix you had but with a few additional changes: https://groups.google.com/g/pflotran-users/c/oHvMZcSAJ0Y?pli=1

lipnikov (approved):

rfiorella:
@jd-moulton reg tests are taking a long time. Should we refactor these to only run on PRs to master or release branches?

rfiorella:
@jd-moulton @ecoon externally-partitioned?

ecoon:
Maybe PARTITION_FROM_EXO to suggest that the partition info is in the exodus file?  It isn't WW-specific -- anyone could use any exodus-based tool to write the variable.  I think exoatt would work, for instance?

Note that under the hood WW just uses (py)metis to partition the surface mesh.

rfiorella:
@ecoon @jd-moulton ready for review now I think.

The mpich issue described above is #933, I suspect this is a pretty small patch at the end of the day but let me know if you think it would be better solved here or in #932.

jd-moulton:
@ecoon For the name I think either your previous suggestion of  "PARTITION_FROM_EXE" or the shorter "FROM_EXO" is good.  The only concern with the shorter is whether some context is lost as it's not hinting at what is coming from the EXO file.  But I'm ok with which ever you prefer. 

@ecoon @rfiorella My opinion on the mpich fix is that since we are introducing patches and updates here already, we should just get it in there too.   I agree the argument that this isn't in our CI testing is a good one however, so if you think that's a determining factor, we can delay it.

rfiorella:
@jd-moulton MSTK patch added, seemed to work in a container and on one Mac I tested.

fmyuan:
I will have a quick test on MPICH patch tomorrow. A first look is good (better than what I patched).

I also made an ugly patch for debugging TPLs' silo. Not sure if want to have it in this PR or not, or have better patch. See: 
https://github.com/fmyuan/amanzi/commit/4cb7816c66ac89ff6dd9fbf15a43e01f3ff47ac1

rfiorella:
@fmyuan thanks for mentioning the HDF5 patch. I've had trouble building debug versions and I wonder if this is why. 

@ecoon and @smolins - David mentioned you both build with debug frequently also. Are you having similar issues? (I suspect the patch from @fmyuan would affect both 0.98.12 and 0.98.13)

lipnikov:
@rfiorella I noticed a failure with the update OSX

cd /Users/lipnikov/ASCEM/build-tpls/build/tpls/superlu_dist/superlu_dist-8.2.1-build/FORTRAN && cp superlu_dist_config.fh /Users/lipnikov/ASCEM/build-tpls/build/tpls/superlu_dist/superlu_dist-8.2.1-build/FORTRAN/.
cp: /Users/lipnikov/ASCEM/build-tpls/build/tpls/superlu_dist/superlu_dist-8.2.1-build/FORTRAN/./superlu_dist_config.fh and superlu_dist_config.fh are identical

lipnikov:
MSTK is broken too

rfiorella:
@lipnikov oh yuck. Ok, the superlu_dist failure is semi-expected, that patch seemed to work once but nowhere else I've tested since. CMakeLists.txt under the Fortran dir of the source seemed to trip an error when compiling tpls RelWithDebInfo and looked like it was expecting source and build dirs to be the same. I'll revert this one until there's a better patch.

The MSTK one is more vexing, as the patch that makes MSTK able to compile with latest mpich breaks the build for openmpi. I know @jd-moulton was hoping to avoid any additional patches after 0.98.13 was accepted, but this is the type of problem I think #932 would be really useful for catching. Wondering if we should revert this patch too and revisit after this PR is merged since it's already quite large.

PR #932, opened 2025-09-23
[WIP] Create matrix of TPL images for broader CI testing

rfiorella:
Adds a matrix of TPL images to test across:
- gcc 11, 14, 15
- 3 versions of mpich
- 3 versions of openmpi 
- 2 versions of cmake (CMAKE_MINIMUM_VERSION in Amanzi and latest release)

Let's merge #930 first and then apply this later on as surely there will be issues across this range of build systems/compilers/MPI.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "ci"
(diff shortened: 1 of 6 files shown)


lipnikov:
@rfiorella Are you suggesting to execute 3x3x3x2 = 54 full builds every time?

rfiorella:
No it's 6, on different runners, so it shouldn't take much if any longer. 

Basically breaks down as an old release, "current" release, and "latest" release sets for gcc, mpi, and cmake.

rfiorella:
Removed gcc 14 as somewhat redundant with gcc:latest. Now builds for each architecture: 4 images spanning two different version "sets" of gcc, MPI, cmake. Runs tests for each architecture, and then stitches individual architecture images together. 

Note however that there will no longer be an image created that has a master-latest tag (instead, it will carry info about the compilers and MPI as well) - should we also designate one of these images to also have a master-latest tag?

PR #944, opened 2025-12-12
WIP: Fixes weird bug in reactive transport tags

ecoon:
Also removes some work-arounds in Amanzi PKs.

WIP because this logic likely needs to be mirrored in Matrix reactive transport.


ecoon:
This ran into some serious difficulties in getting tests to pass.  It needs to be completely revisited and redone.

PR #952, opened 2026-04-10
Add EcoSIM to TPL Collection

jd-moulton:
This pull request is a second revision of the original pull request from @AndrewGraus to address portability and align with the style and variables in Amanzi-ATS's other TPL Build<TPL_NAME>.cmake files.

This PR includes small updates to bootstrap.sh to add the option

--enable-ecosim

for both the TPL and for the EcoSIM PK that is being developed (see ATS PR).   

The EcoSIM build from Amanzi-ATS follows the standard approach used for other TPLs and passes the necessary information to the EcoSIM configuration so that its dependency on HDF5, ZLIB, NETCDF, and NETCDF-fortran, is met with the versions already installed through Amanzi-ATS.

The TPL has been tested on several clusters, including Perlmutter, Lawrencium, smaller local clusters, and is stable with the options included. It is likely that some parts of the templates and options could be cleaned up for efficiency / readability. Particularly with how ATS loads in the EcoSIM libraries.


AndrewGraus:
Quick comment on the last paragraph of the pull request, since the text is from my original pull request. 

@jd-moulton has gone through the CMake build in Amanzi (ATS and EcoSIM as well) and cleaned up the issues I had with it. So I don't think that last part of the pull message applies anymore.

smolins:
@jd-moulton @AndrewGraus Is this ready for approval?  what is the hold?

jd-moulton:
@smolins and @AndrewGraus Initial delay was me thinking this should go at the same time as the ATS PR.   Once I realized that wasn't necessary I was tied up in other things and personal travel.   I will fix the CI and get this merged in the next couple of days.

AndrewGraus:
@jd-moulton @smolins - Jinyun is currently working with LBL to acquire a license for EcoSIM, he says there's a small chance the EcoSIM repo may have to go private by their request temporarily, but it is public now, and will be public after the process is finished. I think we should proceed with the CI and merge, but let me know if that is an issue for you.

smolins:
> @jd-moulton @smolins - Jinyun is currently working with LBL to acquire a license for EcoSIM, he says there's a small chance the EcoSIM repo may have to go private by their request temporarily, but it is public now, and will be public after the process is finished. I think we should proceed with the CI and merge, but let me know if that is an issue for you.

It would be good to know when this will happen. If we are merging the pull request that means that the source for ecosim needs to be available, or are we creating a container with all the built TPLs, so it won't matter but for folks trying to install the TPLs with EcoSIM (unlikely scenario)?

In case, not sure why the CI failed. I can only see that last time it was triggered it was cancelled after a few seconds. So can't understand the error.

jd-moulton:
@smolins I agree this is a bit awkward if it's temporarily unavailable.  You're right that if we get it into the Docker image then it won't affect the CI or people building within Docker.  And most people aren't ready to use it yet.  I tend to say that we should push forward and if it really goes off line for a bit we try to for it to be disabled or make the option in bootstrap.sh "smart" enough to only turn on when the EcoSIM TPL is available.

smolins:
@jd-moulton @AndrewGraus who wants to do the honors?

jd-moulton:
I was planning to do it when I get out of my meetings today.   After the merge, we need to check the github actions on master (as they'll likely pick up changes from here that we don't want), and then it would be good to rebuild the TPLs in Docker from the master branch to make sure all is good there.

smolins:
Ok, thanks for the update. I would not know how to generate the docker
containers for the tpls build.
Sergi

jd-moulton:
@smolins There was as time when it was difficult to generate the multi-arch containers for the TPLs.   But after struggling with it for the ATS short course I created a build-tpls.yml action that can be activated manually.  So now it is pretty straightforward.  I will check, we should make sure this is covered in the documentation.

PR #956, opened 2026-05-29
Use one rank to read file data and broadcast 

gaobhub:
I got "double free or corruption" error with the "independent variable from file" evaluator when using multiple cores under amanzi-1.5. It seems because HDF5_MPI::readFieldData_ requires all ranks to simultaneously open and read the same .h5 file. Revised to using one rank to read and then broadcast to other ranks. I did not get errors with amanzi-1.6. It seems amanzi-1.6 should have the same issue but somehow it is not triggered.

### DIFF
--- a/src/state/evaluators/EvaluatorIndependentFromFile.cc
+++ b/src/state/evaluators/EvaluatorIndependentFromFile.cc
@@ -105,24 +105,35 @@ EvaluatorIndependentFromFile::EnsureCompatibility(State& S)
 
   // load times, ensure file is valid
   // if there exists no times, default value is set to +infinity
-  HDF5Reader reader(filename_);
   times_.clear();
+  int rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
   if (temporally_variable_) {
-    try {
-      reader.ReadData("/time", times_);
-    } catch (...) {
-      std::stringstream messagestream;
-      messagestream << "Variable " << my_key_ << " is defined as a field changing in time.\n"
-                    << " Dataset /time is not provided in file " << filename_ << "\n";
-      Errors::Message message(messagestream.str());
-      Exceptions::amanzi_throw(message);
+    if (rank == 0) {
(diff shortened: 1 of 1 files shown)


lipnikov:
HDF5_MPI::readFieldData_ is implemented for distributed vectors: each rank reads the portion corresponding to its local Epetra_Vector. Potentially, the problem you observed was triggered by something else.

gaobhub:
> HDF5_MPI::readFieldData_ is implemented for distributed vectors: each rank reads the portion corresponding to its local Epetra_Vector. Potentially, the problem you observed was triggered by something else.

Probably, but with single rank for file read and broadcast solved my issue. I am not sure what exactly caused my heap corruption error.

lipnikov:
Do you have a direct reproducer or can you describe a serious of steps which lead to this problem?

gaobhub:
> Do you have a direct reproducer or can you describe a serious of steps which lead to this problem?

OK, let me describe it. I used the "independent variable from file" evaluator to read a hdf5 file, which includes values for surface cells (one value for each cell, 1000 in total) at time=0 (constant in time: true). I used the data from this file to multiply another variable output by model at surface cells. I used 32 cores to run the model. I got errors 
"
double free or corruption (out); 
free(): invalid next size (fast); 
...........
backtrace:
....../xlib/libstate.so(_ZN6Amanzi28EvaluatorIndependentFromFile9LoadFile_Ei+0x295)
"
There was no error with one core, but the same error came out if using more than one core.

First, I revised the "EnsureCompatibility" part. I moved "HDF5Reader reader(filename_)" inside "if (rank == 0)" instead applying it to all ranks. Crush disappeared but the model hanged there reading file without any progress. Then, I revised to use one rank also in "EvaluatorIndependentFromFile::LoadFile_(int i)" and broadcast. It works finally.

lipnikov:
Amanzi has a few tests in src/mpc which use "independent variable from file'" evaluator. Since all tests pass, I can only suggest to look first at this tests.

PR #969, opened 2026-08-03
added an evaluator to allow for generating tensor from input file

gaobhub:


src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
```diff
@@ -0,0 +1,334 @@
+/*
+  Copyright 2010-202x held jointly by participating institutions.
+  Amanzi is released under the three-clause BSD License.
+  The terms of use and "as is" disclaimer for this license are
+  provided in the top-level COPYRIGHT file.
+
+  Authors: Ethan Coon, Bo Gao
```
ecoon, line 7: I didn't write this...

src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
```diff
@@ -0,0 +1,334 @@
...
+EvaluatorIndependentTensorFromFile::EnsureCompatibility(State& S)
+{
+  auto& f = S.Require<TensorVector, TensorVector_Factory>(my_key_, my_tag_, my_key_);
+ 
+  if (rank_ == -1 && f.map().Mesh().get()) {
+    dimension_ = f.dimension();
+    AMANZI_ASSERT(dimension_ > 0);
+ 
+    tensor_type_ = plist_.get<std::string>("tensor type");
+    if (tensor_type_ == "scalar") {
```
ecoon, line 107: We should probably make this snippet of code a helper function in TensorVector or something.  Don't worry about it for now, just noting it.

src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
```diff
@@ -0,0 +1,334 @@
...
+        times_ = times.toVector();
+      } catch (...) {
+        std::stringstream messagestream;
+        messagestream << "Variable " << my_key_ << " is defined as a field changing in time.\n"
+                      << " Dataset /time is not provided in file " << filename_ << "\n";
+        Errors::Message message(messagestream.str());
+        Exceptions::amanzi_throw(message);
+      }
+    } else {
+      times_.push_back(1e+99);
```
ecoon: Please use:

std::numeric_limits<double>::max

src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
```diff
@@ -0,0 +1,334 @@
...
+    for (int j = 1; j < times_.size(); ++j) {
+      if (times_[j] <= times_[j - 1]) {
+        Errors::Message m;
+        m << "EvaluatorIndependentTensorFromFile: time values are not strictly increasing";
+        throw(m);
+      }
+    }
+ 
+    current_interval_ = -1;
+    t_before_ = -1.0e+99;
```
ecoon: and 

std::numeric_limits<double>::lowest()

src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
```diff
@@ -0,0 +1,334 @@
...
+ 
+  double t = S.get_time(my_tag_);
+  if (time_func_ != Teuchos::null) {
+    std::vector<double> point(1, t);
+    t = (*time_func_)(point);
+  }
+ 
+  // Check if we are before the current interval
+  if (t < t_before_) {
+    t_before_ = -1.0e+99;
```
ecoon: std::numeric_limits<double>::lowest()

src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
```diff
@@ -0,0 +1,334 @@
...
+  if (t < t_before_) {
+    // should never be possible thanks to the previous check
+    AMANZI_ASSERT(0);
+  } else if (t == t_before_) {
+    // at the start of the interval
+    AMANZI_ASSERT(val_before_ != Teuchos::null);
+    cv_interp = *val_before_;
+
+  } else if (t < t_after_) {
+    if (t_before_ == -1.0e+99) {
```
ecoon: std::numeric_limits<double>::lowest()

src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
```diff
@@ -0,0 +1,334 @@
...
+    cv_interp = *val_after_;
+
+  } else {
+    // to the right of the interval -- advance the interval
+    while (t > t_after_) {
+      current_interval_++;
+      t_before_ = t_after_;
+      if (current_interval_ + 1 == times_.size()) {
+        // at the end of data
+        t_after_ = 1.0e99;
```
ecoon: std::numeric_limits<double>::max()

ecoon (changes requested):
I'd prefer to get rid of the code duplication in Update_, LoadFile_, and Interpolate_.

Can you make a set of helper functions, in a shared file e.g. EvaluatorFromFile_Helpers.hh:

```
namespace Amanzi {
namespace State {
namespace EvaluatorFromFile_Helpers {
template<Evaluator_type> void LoadFile(int i, Evaluator_type& eval) {...}```

}
}
}
```


Then you'll have to friend the function in the evaluator so that the function can access private data:
```
#include "EvaluatorFromFile_Helpers.hh"
class EvaluatorIndependentTensorFromFile {
...
friend EvaluatorFromFile_Helpers::LoadFile_<EvaluatorIndependentTensorFromFile>;
};
```

Then you can get rid of the methods for LoadFile and Interpolate, and the Update method can call the Update function (which takes a CompositeVector) directly and map that CV into the TV.

Then update the EvaluatorIndependentFromFile to use these functions as well.

This way we just have one place for these, which I think is particularly useful for the Update which does some complicated logic.

src/state/evaluators/EvaluatorIndependentTensorFromFile.cc
gaobhub: I copied you code.

src/state/evaluators/EvaluatorIndependentTensorFromFile.hh
```diff
@@ -0,0 +1,145 @@
...
+  // ---------------------------------------------------------------------------
+  // Constructors
+  // ---------------------------------------------------------------------------
+  explicit EvaluatorIndependentTensorFromFile(Teuchos::ParameterList& plist);
+
+  EvaluatorIndependentTensorFromFile(const EvaluatorIndependentTensorFromFile& other) = default;
+
+  virtual Teuchos::RCP<Evaluator> Clone() const override;
+
+  virtual Evaluator& operator=(const Evaluator& other) override;
```
lipnikov, line 102: Do we need this member functions in this class?

gaobhub: Evaluator.hh has this `virtual Evaluator& operator=(const Evaluator& other) = 0; `, so probably we also need to include it?

src/state/evaluators/EvaluatorFromFile_Helpers.hh
```diff
@@ -0,0 +1,211 @@
...
+}
+
+
+inline void Interpolate(double time, double t_before, double t_after,
+                        const CompositeVector& val_before,
+                        const CompositeVector& val_after,
+                        CompositeVector& v)
+{
+  AMANZI_ASSERT(t_before >= 0.0);
+  AMANZI_ASSERT(t_after >= 0.0);
```
lipnikov, line 65: this check follows from other checks, so we can remove it.

src/state/evaluators/EvaluatorIndependentTensorFromFile.hh
lipnikov: You are correct.

src/state/evaluators/EvaluatorFromFile_Helpers.hh
```diff
@@ -0,0 +1,211 @@
...
+          Interpolate(t, eval.t_before_, eval.t_after_,
+                      *eval.val_before_, *eval.val_after_, cv);
+        }
+      }
+    }
+  }
+}
+
+
+inline void CopyVectorToTensorVector(const Epetra_MultiVector& v, int j, TensorVector& tv)
```
ecoon, line 145: Isn't this implemented elsewhere?  And I think with a bit more generality?  I'll go looking.

ecoon (changes requested):
I'll take a look at this... I think there is some shared code elsewhere that we can avoid replicating.

src/state/evaluators/EvaluatorFromFile_Helpers.hh
gaobhub: This was originally in EvaluatorIndependentTensorFunction.cc. I moved it to EvaluatorFromFile_Helpers.hh to share it with EvaluatorIndependentTensorFunction.cc and EvaluatorIndependentTensorFromFile.cc.

PR #970, opened 2026-08-20
Update primary_ion_concentration with timestep

gaobhub:
Two main things were changed:
(1) Add CommitStep for primary_ion_concentration to update it with timestep;
(2) Allow Alquimia pk to accept checkpoint file as initial condition.

### DIFF
--- a/src/pks/chemistry/Alquimia_PK.cc
+++ b/src/pks/chemistry/Alquimia_PK.cc
@@ -38,6 +38,7 @@
 
 // Chemistry
 #include "Alquimia_PK.hh"
+#include "EvaluatorPrimary.hh"
 
 namespace Amanzi {
 namespace AmanziChemistry {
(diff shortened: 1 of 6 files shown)


src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -965,5 +1004,42 @@ Alquimia_PK::copyFields_(const Tag& tag_dest, const Tag& tag_source)
...
+Alquimia_PK::CommitStep(double t_old, double t_new, const Tag& tag_next)
+{
+  Chemistry_PK::CommitStep(t_old, t_new, tag_next);
+
+  Tag tag_current = tag_next == tag_next_ ? tag_current_ : Tags::CURRENT;
+
+  // Copy primary_free_ion_concentration from new state at tag_next to tag_current, 
+  // then at the next timestep, transport ats pk can use the updated value.
+  if (!primary_ion_conc_key_.empty()) {  
+    assign(primary_ion_conc_key_, tag_current, tag_next, *S_);
```
ecoon, line 1017: I feel like there is a helper method in PK_Helpers that does BOTH the assign AND the SetChanged() call for you.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -369,7 +371,11 @@ Alquimia_PK::Setup()
       .SetMesh(mesh_)
       ->SetGhosted(false)
       ->SetComponent("cell", AmanziMesh::Entity_kind::CELL, aux_out_subfield_names_[name].size());
-    S_->GetRecordW(key, tag_next_, passwd_).set_io_checkpoint(false);
+
+    if (name == "primary_free_ion_concentration") {
```
ecoon, line 375: I don't think you want to do this -- this requires that you put a "primary variable" evaluator in the input file, right?

Instead, just use:

PK_Helpers::requireEvaluatorAssign()

which is intended to do this for you.

src/pks/chemistry/Alquimia_PK.cc
```diff
@@ -389,14 +395,47 @@ Alquimia_PK::Initialize()
...
+    restart_filename = plist_->sublist("initial conditions").get<std::string>("restart file");
+  }
+
+  updateSubstate(Tags::DEFAULT);
+
+  // initialized aux_data and aux_out from restart file if there is
   for (const auto& [name, key] : aux_out_names_) {
-    S_->GetRecordW(key, tag_next_, passwd_).set_initialized();
+    Record& rec = S_->GetRecordW(key, tag_next_, passwd_);
+    if (restarted) {
```
ecoon, line 411: I don't think you should have to do this?  What does restarted mean?  I think it should correctly pick up from the restart file on a restarted run?

src/pks/chemistry/Alquimia_PK.cc
ecoon: Right -- if you use requireEvaluatorAssign() above, then you don't need any of this and the call to assign() will also call SetChanged for you.

ecoon (changes requested):

src/pks/chemistry/Alquimia_PK.cc
gaobhub: No, restarted means there is a restart file for initialization. If we do not add this, the code won't pick up data from user provided initialization file, but use the initial condition from pflotran by Alquimia_PK::initializeSingleCell_()

`  if (plist_->isSublist("initial conditions") && 
    plist_->sublist("initial conditions").isParameter("restart file")) {
    restarted = true;
    restart_filename = plist_->sublist("initial conditions").get<std::string>("restart file");
  }
`

src/pks/chemistry/Alquimia_PK.cc
gaobhub: I see, revised.

src/pks/chemistry/Alquimia_PK.cc
gaobhub: I used requireEvaluatorAtCurrent() which seems to be able to do more than requireEvaluatorAssign() and maybe more powerful for any future changes? I am not sure, but I tried to change it to requireEvaluatorAssign(), it works. Should we use requireEvaluatorAssign() here?

PR #972, opened 2026-09-11
Changes to how Operators deal with Symbolic Assembly

### LINKED ISSUE #971: Some permafrost models get timestep crash in commit e91f3bf
Some changes introduced in commit e91f3bf0b45a9e72aea6427b803ddba6212909c1, possibly in $AMANZI_SRC_DIR/src/operators/TreeOperator.cc, appear to cause timestep crashes for some permafrost models (e.g., ats_regression_test-08_energy-permafrost-wrm_vangenuchten-krel_vangenuchten). Andrew (AGraus@lbl.gov) did a test and found that the crash occurs on arm64 architectures but not on x86 architectures.

ecoon:
Commit e91f3bf introduced a bug in forcing a re-symbolic assemble of a TreeOperator in Init, which is intended to only change values.  This led to a non-obvious bug where a new matrix was created but not actually used by the preconditioner in an InverseAssembled object.  The new matrix WAS used in the forward operator of InverseIterative, so often GMRES would "clean up" the lagged preconditioner (but not always!)

This makes three changes with regards to Operator and TreeOperator:

1. Fixes the bug by removing the reset of symbolic assembly in Init
2. Adds a mechanism whereby AssembleMatrix() will call SymbolicAssembleMatrix() if it has not yet been done.
3. Operator may now only be symbolic assembled once -- changing the structure of an operator after symbolic assembly has been done will now error.  This avoids developer errors in creating the new operator but forgetting to deal with Inverses or other objects that still have shared ownership of the underlying matrix.  It also matches the original design intent -- set the structure, then symbolic assemble, then repeatedly assemble and apply, but don't change the structure.

Note that developers CAN still Clone() an operator and then change the structure of the newly constructed operator -- cloning creates a non-symbolic-assembled object.

Closes #971

### DIFF
--- a/src/operators/TreeOperator.cc
+++ b/src/operators/TreeOperator.cc
@@ -50,7 +50,7 @@ TreeOperator::TreeOperator()
     initialize_complete_(false),
     compute_complete_(false),
     assembly_complete_(false),
-    symbolic_assembly_complete_(false)
+    structure_locked_(false)
 {
   vo_ = Teuchos::rcp(new VerboseObject("TreeOperator", Teuchos::ParameterList()));
 }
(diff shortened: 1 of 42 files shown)


src/operators/TreeOperator.cc
```diff
@@ -130,10 +130,44 @@ TreeOperator::TreeOperator(const TreeOperator& other)
...
+* Locks the structure of this TreeOperator and, recursively, every
+* sub-TreeOperator block and leaf Operator beneath it.
+****************************************************************** */
+void
+TreeOperator::lockStructure()
+{
+  structure_locked_ = true;
+  for (int i = 0; i != row_size_; ++i) {
+    for (int j = 0; j != col_size_; ++j) {
+      if (blocks_[i][j] != Teuchos::null) blocks_[i][j]->lockStructure();
```
lipnikov, line 160: Should we instead check that the block-structure was locked? The current design forces to lock the block structure in PKs quite earlier which we probably do already.

lipnikov (approved):

ecoon:
@lipnikov  there is some Amanzi code that is throwing Epetra errors that are silently ignored.  Specifically, the H2O_DensityIAPWS97.hh model for density has dependencies on some weird stuff that I'm not sure I understand (physically).  Structurally it looks like there is an Epetra error where chain rule is attempted where density is a function of some 18-component vector, and the code errors when trying to multiply through some chain rule calculations there.

I suspect you want to explicitly turn off some partial derivatives.  I don't understand the model well enough to do so myself.

mpc_supercriticalPH 

will print these errors if you configure Trilinos to print errors (or to treat Epetra errors as fatal).


(This is unrelated to this PR, but I found it debugging the last failed CI....)

ecoon:
One more potential "gotcha" here @lipnikov .  Operator::Clone() now shares the same RHS as the cloned operator.  This was necessary for a Amanzi MPC tests that must call clone.  

So say the developer calls:

B = A.Clone()

I think some Amanzi MPCs may populate B's RHS vector through, e.g., a PDE_Accumulation that was constructed to use A.

The problem is what happens if someone adds to B's Ops by adding another term that uses the RHS -- now A will also see the changes made to B's RHS.

Maybe this is ok?  I'm not sure what the right thing to do here is, but for now tests pass...

lipnikov:
Thank you for two error reports. The first one is confirmed. I'm looking at it... 

MORE: I fixed the error by introducing the evaluator with detached dependent variables. BTW, Epetra also issues "Epetra Error:  Warning in Epetra_Import: Target IDs not found in Source Map (Do you want to import to subset of Target Map?)  Error Code:  1" for parallel runs.

lipnikov:
I think that the RHS should be local. I'll revisit the logic of failed MPCs to see why the RHS value is needed after cloning. Thanks!

PR #974, opened 2026-09-17
Subscript operator was added to struct View_iter

lipnikov:
Subscript operator was added to struct View_iter. This closes the error message "'Kokkos::Impl::View_iter<Kokkos::MeshView<int *, Kokkos::HostSpace>>' does not provide a subscript operator"

Details: The newer standard library implementation uses the subscript operator (__first[__start]) internally to swap or shift elements. Because Kokkos::Impl::View_iter has never explicitly provided operator[], the compiler errors.


JulienLoiseau (approved):
You could use `KOKKOS_INLINE_FUNCTION` like for the other members. 

Also there might be a bug in the `operator-`, it should create an iterator and return it not change the current one I think with: 

```
KOKKOS_INLINE_FUNCTION View_iter operator-(const int& decr) const
{
  View_iter tmp(*this);
  tmp -= decr;
  return tmp;
}
```

lipnikov:
@JulienLoiseau I removed operator- due to ambiguity with the earlier defined operator-.

ecoon:
This probably fixes an issue I just ran against and opened a new PR for.  Looks very close, but I have a few more tweaks.  Will move my branch onto this one.

ecoon (approved):

JulienLoiseau (approved):

PR #975, opened 2026-09-23
fix to Iterators::operator[]

ecoon:
new clang/standard c++ library identified an incorrect implementation in Iterators that now cause an error


ecoon:
duplicate to #974

